PR #4, opened 2018-07-16
Observation Accessors for:

ahubley:
AutopsyPerformed
	AutopsyResultsAvailable
	MannerOfDeath
	TobaccoUseContributedToDeath

Added conditions with no onset for Contributing Conditions in output

### DIFF
--- a/FhirDeathRecord/DeathRecord.cs
+++ b/FhirDeathRecord/DeathRecord.cs
@@ -1,5 +1,7 @@
+using System;
 using System.Linq;
 using System.Text.RegularExpressions;
+using System.IO;
 using Hl7.Fhir.ElementModel;
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Serialization;
@@ -45,11 +47,12 @@ namespace FhirDeathRecord
             var matches = Navigator.Select(path);
             if (matches.Count() > 0)
             {
-                return (string)matches.First().Value;
+                return Convert.ToString(matches.First().Value);
             }
             else
             {
                 // TODO: Figure out appropriate way to handle element not found
+                Console.WriteLine("element not found");
                 return "";
             }
         }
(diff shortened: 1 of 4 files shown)


FhirDeathRecord/DeathRecord.cs
```diff
@@ -45,11 +47,12 @@ private string GetFirst(string path)
             var matches = Navigator.Select(path);
             if (matches.Count() > 0)
             {
-                return (string)matches.First().Value;
+                return Convert.ToString(matches.First().Value);
             }
             else
             {
                 // TODO: Figure out appropriate way to handle element not found
+                Console.WriteLine("element not found");
```
holmesie: I've thought a bit more about this case. I think (and am doing this locally right now with my ongoing additions) we should return `null` here. My thoughts are:
- It seems easier, or perhaps more clear, to check for `null` inside any app that wants to use this code, when checking to see if a "thing" exists in a record. We shouldn't be doing anything special, or making **any** assumptions on behalf of any users of this library.
- We shouldn't be printing anything to the console.
- In our CLI app, we print whatever these functions return. The output is the same for `null` and `""` (Console.WriteLine outputs nothing for both).

For example, `DeathRecord.AutopsyPerformed()` returning `null` if the record does not have an entry for Autopsy Performed seems the most natural thing to me.

Pinging @pkmitre for his thoughts on this as well, as the TODO was his original comment.

FhirDeathRecord/DeathRecord.cs
```diff
@@ -84,6 +87,42 @@ public string ContributingConditions
                 return htmlRegex.Replace(cc, "");
             }
         }
+
+        public string AutopsyPerformed
+        {
+            get
+            {
+                string ap = GetFirst("Bundle.entry.resource.where($this is Observation).where(code.coding.code='85699-7').value");
+                return ap;
```
holmesie: These should probably just be one line, like:
```
return GetFirst("Bundle.entry.resource.where($this is Observation).where(code.coding.code='85699-7').value");
```

FhirDeathRecord/DeathRecord.cs
pkmitre: I tried to think a bit about what it would look like if we raised an error; that doesn't seem right, since we couldn't really bubble it up to the user because the user needs to be able to ignore missing fields. `null` seems reasonable to me; one reason not to do that would be if the (IJE-inspired) receiving interface expects a string, but that seems likely easy to deal with.

FhirDeathRecord/DeathRecord.cs
```diff
@@ -45,11 +47,12 @@ private string GetFirst(string path)
             var matches = Navigator.Select(path);
             if (matches.Count() > 0)
             {
-                return (string)matches.First().Value;
+                return Convert.ToString(matches.First().Value);
```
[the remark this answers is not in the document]
ahubley: I believe so. Typecasting like `(string)variable` fails on booleans, so I switched it to use the Convert package.

PR #95, opened 2020-02-26
Addresses issues with the mapping to the internal format used by nightingale

pkmitre:
This pull request makes the following changes:

* Allows the certificate number to be represented
* Adds contributing causes
* Fixes an issue with disposition location mapping
* Fixes an issue where the tests were not running correctly

### DIFF
--- a/VRDR.HTTP/Nightingale.cs
+++ b/VRDR.HTTP/Nightingale.cs
@@ -59,7 +59,7 @@ namespace VRDR.HTTP
             SetStringValueDictionary(values, "cod.under3", record.COD1D);
             SetStringValueDictionary(values, "cod.under3Int", record.INTERVAL1D);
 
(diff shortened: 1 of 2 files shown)


VRDR.HTTP/Nightingale.cs
```diff
@@ -463,36 +473,35 @@ public static DeathRecord FromNightingale(string json)
...
-                case "Divorced (but not remarried)":
+                case "Divorced":
                     mar = new Dictionary<string, string>();
                     mar.Add("code", "D");
                     mar.Add("system", "http://terminology.hl7.org/CodeSystem/v3-MaritalStatus");
                     mar.Add("display", "Divorced");
                     deathRecord.MaritalStatus = mar;
                     break;
-                case "Never married":
+                case "Never Married":
```
holmesie: Good catch on these, but maybe it would have been better to have originally made the switch param `.ToLower`, and have these lower case.

PR #104, opened 2020-04-22
Bump the versions of netcoreapp for the VRDR.CLI and VRDR.Tests.

rbclark:
Previously had both of those projects were still using netcoreapp2.1 even though we upgraded everything else to dotnet 3.1.

This was causing the following error on a clean install of dotnet 3.1:

```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '2.1.0' was not found.
  - The following frameworks were found:
      3.1.3 at [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=2.1.0&arch=x64&rid=osx.10.15-x64
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Nuget](https://img.shields.io/nuget/v/VRDR.Messaging?label=VRDR.Messaging%20%28nuget%29)](https://www.nuget.org/packages/VRDR.Messaging)
 
 # vrdr-dotnet
-This repository includes .NET (C#) code for 
+This repository includes .NET (C#) code for
 
 - Producing and consuming the Vital Records Death Reporting (VRDR) Health Level 7 (HL7) Fast Healthcare Interoperability Resources (FHIR) standard. [Click here to view the FHIR Implementation Guide](http://hl7.org/fhir/us/vrdr/2019May/).
 - Producing and consuming FHIR messages for the exchange of VRDR documents.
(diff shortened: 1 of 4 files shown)


rbclark:
Since we are using vrdr-dotnet with canary which is on dotnet 3.1 it may be beneficial to test on both, I think I have that working in my latest commit. We can also Multi-Target VRDR.CLI without any issues it seems, I can remove that and we can go to a specific version if we want to.

README.md
```diff
@@ -336,40 +336,42 @@ dotnet test VRDR.Tests/DeathRecord.Tests.csproj
 ### VRDR.CLI
 This directory contains a sample command line interface app that uses the VRDR library to do a few different things.
 
+NOTE: In all of the below commands you must specify your preferred version of dotnet, either netcoreapp2.1 or netcoreapp3.1, depending on whether you have installed dotnet 2.1 or dotnet 3.1 locally on your system.
```
hadleynet, line 339: Is there any way we can set default so the currently installed runtime will be used? It seems like a step back to require this extra flag every time.

rbclark: I messed with it a bit and it I think I have found a solution. I set `<TargetFramework>` to netcoreapp3.1 and then set put both frameworks in `<TargetFrameworks>`. This has allowed me to run it on dotnet core 3.1 without a `--framework` flag, however running on dotnet core 2.1 requires a flag. I unfortunately can't make it completely seamless based on what versions the user has installed.

PR #106, opened 2020-04-24
Merge MortalityData.cs with the one from canary, add tests

rbclark:
The MortalityData.cs in canary was rewritten without realizing that it was the same as the one in VRDR. This PR ports the updated version back over to VRDR and moves the tests over. Another PR will remove it from canary and reference this file instead.


rbclark:
I'm definitely fine with this, do we know where this data came from? If this already came from a JSON file that someone else maintains it we can just reference that.

PR #107, opened 2020-04-28
Improve error handling, add helper class for entity axis coding

hadleynet:


VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -311,6 +311,78 @@ public CauseOfDeathEntityAxisEntry(string deathCertificateText, string causeOfDe
...
+        /// <summary>
+        /// Build a List&lt;CauseOfDeathEntityAxis&gt; from the currently contained set of codes.
+        /// </summary>
+        /// <returns>cause of death entity axis coding list</returns>
+        public List<CauseOfDeathEntityAxisEntry> ToCauseOfDeathEntityAxis()
+        {
+            var list = new List<CauseOfDeathEntityAxisEntry>();
+            foreach (KeyValuePair<int, SortedDictionary<int, string>> pair in codes)
+            {
+                var entry = new CauseOfDeathEntityAxisEntry("", pair.Key.ToString());
```
pkmitre, line 339: For clarity, perhaps putting `pair.Key.ToString()` into a temporary `lineNumber` variable would be helpful to indicate what the data is? Not sure about what you'd call `pair.Value.Values` if it made sense to do the same for that...

VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -311,6 +311,78 @@ public CauseOfDeathEntityAxisEntry(string deathCertificateText, string causeOfDe
...
+        /// <item>Part I. Line c</item>
+        /// <item>Part I. Line d</item>
+        /// <item>Part I. Line e</item>
+        /// <item>Part II</item>
+        /// </list>
+        /// </param>
+        /// <param name="position">Sequence within line</param>
+        /// <param name="code">ICD code</param>
+        /// <exception cref="System.ArgumentException">Thrown if <c>line</c> or <c>position</c> is not a number</exception>
+        public void Add(string line, string position, string code)
```
pkmitre, line 367: Maybe `lineNumber` would be more explicit than `line`? (And then `lineNumberValue` below)

VRDR.Messaging/DeathRecordSubmission.cs
```diff
@@ -42,9 +42,12 @@ public DeathRecord DeathRecord
             {
                 deathRecord = value;
                 MessageBundle.Entry.RemoveAll( entry => entry.Resource.ResourceType == ResourceType.Bundle );
-                MessageBundle.AddResourceEntry(deathRecord.GetBundle(), "urn:uuid:" + deathRecord.GetBundle().Id);
                 Header.Focus.Clear();
-                Header.Focus.Add(new ResourceReference(deathRecord.GetBundle().Id));
+                if (deathRecord != null)
+                {
+                    MessageBundle.AddResourceEntry(deathRecord.GetBundle(), "urn:uuid:" + deathRecord.GetBundle().Id);
+                    Header.Focus.Add(new ResourceReference(deathRecord.GetBundle().Id));
```
pkmitre, line 49: I'll confess I don't fully understand internal reference rules... should urn:uuid: be used here since focus is a reference? Does `ResourceReference` take care of that?

VRDR.Tests/Messaging_Should.cs
```diff
@@ -26,29 +26,48 @@ public void CreateSubmission()
         {
             DeathRecordSubmission submission = new DeathRecordSubmission();
             Assert.Equal("http://nchs.cdc.gov/vrdr_submission", submission.MessageType);
+            Assert.Null(submission.DeathRecord);
+            Assert.Equal("http://nchs.cdc.gov/vrdr_submission", submission.MessageDestination);
+            Assert.NotNull(submission.MessageTimestamp);
+            Assert.Equal("nightingale", submission.MessageSource);
```
pkmitre, line 32: I'll confess I didn't notice this before... I'm not sure that nightingale is a good default for the source endpoint (or that there should be a default for the jurisdiction side of these messages).

VRDR.Messaging/AckMessage.cs
```diff
@@ -5,10 +5,12 @@ namespace VRDR
     /// <summary>Class <c>AckMessage</c> supports the acknowledgment of other messages.</summary>
     public class AckMessage : BaseMessage
     {
+        private const string ACK_MESSAGE = "http://nchs.cdc.gov/vrdr_acknowledgement";
```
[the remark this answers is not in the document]
hadleynet: OK. I think it was used in two places but then I refactored to remove one of them.

VRDR.Messaging/CodingResponseMessage.cs
hadleynet: OK.

VRDR.Messaging/CodingResponseMessage.cs
hadleynet: OK.

VRDR.Messaging/DeathRecordSubmission.cs
hadleynet: Good catch!

VRDR.Tests/Messaging_Should.cs
hadleynet: OK, I can remove this.

hadleynet:
I have marked the CauseCodes class as obsolete (deprecated) in preparation for removal in the following release.

hadleynet:
I added text to the README to cover the new helper class but also left in the existing text covering the entity axis API since there are two ways of using the API that may be useful in different circumstances.

PR #110, opened 2020-05-11
Add Business Identifiers to Every Message

hadleynet:
This branch will bring the VRDR.Messaging library in compliance with the 5/8/2020 version of the messaging proposal specification.


VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -283,31 +241,32 @@ public List<CauseOfDeathEntityAxisEntry> CauseOfDeathEntityAxis
...
+        /// <summary>The line identifier of the corresponding cause of death in the death certificate</summary>
+        /// <list type="number">
+        /// <item>Part I. Line a</item>
+        /// <item>Part I. Line b</item>
+        /// <item>Part I. Line c</item>
+        /// <item>Part I. Line d</item>
+        /// <item>Part I. Line e</item>
+        /// <item>Part II</item>
+        /// </list>
+        public readonly string LineNumber;
```
pkmitre, line 253: Perhaps I'm misunderstanding how this documentation works, but I don't see how this &lt;list /&gt; applies to LineNumber

VRDR.Messaging/BaseMessage.cs
```diff
@@ -192,6 +209,73 @@ public string MessageDestination
...
+                }
+            }
+        }
+
+        /// <summary>Create an NCHS identifier for the supplied DeathRecord.</summary>
+        /// <param name="record">the DeathRecord from which the year of death, jurisdiction of death and certificate id will be extracted.</param>
+        /// <returns>the NCHS compound identifier for the supplied DeathRecord.</returns>
+        protected static string CreateNCHSIdentifier(DeathRecord record)
+        {
+            var ije = new IJEMortality(record);
```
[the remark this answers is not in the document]
hadleynet: There's quite a bit of logic in the IJE class to handle stuff like padding etc that is shared over multiple IJE fields. I think this logic belongs in the IJE class since it is related to the IJE fixed length field format which the NCHSIndetifier property inherits. Duplicating that code elsewhere could result in maintenance issues.

VRDR.Messaging/CodingResponseMessage.cs
hadleynet: Each item is a value of the line number parameter. E.g. if LineNumber is 1, that represents Part I. Line a (as per the TRANSAX format). I did consider adding an enum to make this even clearer?

PR #111, opened 2020-05-20
Adding task-oriented documentation for use of the messaging library

hadleynet:


VRDR.Messaging/BaseMessage.cs
```diff
@@ -76,31 +76,35 @@ public BaseMessage(String messageType)
         }
 
         /// <summary>Helper method to return a XML string representation of this DeathRecordSubmission.</summary>
+        /// <param name="prettyPrint">controls whether the returned string is formatted for human readability (true) or compact (false)</param>
```
pkmitre, line 79: Seems a bit odd to add pretty printing functionality here, but not in other parts of the library, what's the goal?

doc/Messaging.md
```diff
@@ -0,0 +1,219 @@
...
+
+If extraction was succesful, NCHS can generate an acknowledgement message and send it to the submitting jurisdiction. An `AckMessage` constructor is available that accepts a source message parameter (`submission` in the example below) and this is used to initialize `AckMessage` properties:
+
+- `AckMessage.MessageDestination` will be assigned the value of `source.MessageSource`
+- `AckMessage.AckedMessageId` will be assigned the value of `source.MessageId`
+- `StateIdentifier` will be assigned the value of `source.StateIdentifier`
+- `CertificateNumber` will be assigned the value of `source.CertificateNumber`
+- `NCHSIdentifier` will be assigned the value of `source.NCHSIdentifier`
+
+The constructor will also default the value of the `MessageSource` property to `http://nchs.cdc.gov/vrdr_submission` which is suitable for messaging originating from NCHS.
```
pkmitre, line 65: Wondering if this default makes sense given the jurisdiction use of Acks (as you accurately note below)... can we populate the MessageSource property from the MessageDestination of the message being acknowledged? My intuition is no because that inappropriately trusts the contents of message being acknowledged...

doc/Messaging.md
```diff
@@ -0,0 +1,219 @@
...
+### Acknowledge Coding
+
+If extraction of the coding information was succesful, the jurisdiction can generate an acknowledgement message and send it to NCHS. As described earlier, the `AckMessage` constructor initializes properties based on the source coding message.
+
+The constructor will default the value of the `MessageSource` property to `http://nchs.cdc.gov/vrdr_submission` which is not suitable for messages originating from jurisdictions so this needs to be overriden.
+
+```cs
+// Create the acknowledgement message
+var ackMsg = new AckMessage(coding);
+ackMsg.MessageSource = "https://example.com/juristdiction/message/endpoint";
```
pkmitre, line 190: Is setting MessageSource outside the constructor just to make it explicit to the reader?

VRDR.Messaging/BaseMessage.cs
hadleynet: Goal was to make it easier to create human-readable example messages for testing. Currently have to find a separate JSON/XML formatter, to pretty print the condensed JSON/XML output.

doc/Messaging.md
hadleynet: Nice idea. On balance I think defaulting the ack source to the acked message destination is probably a better default that the current fixed one.

doc/Messaging.md
hadleynet: That was the idea. If we make the change above I'd remove this though.

VRDR.Messaging/BaseMessage.cs
pkmitre: If you don't already use it I recommend jq

doc/Messaging.md
pkmitre: I think the last bit is tweaking this example to match the code changes and then we're good to merge

PR #116, opened 2020-06-24
Convert ExaminerContacted backend from FHIRBoolean to CodeableConcept

rbclark:
This was updated in R4 however the change did not make it into the initial migration of the library.

https://trifolia-fhir.lantanagroup.com/igs/lantana_hapi_r4/vrdr/StructureDefinition-VRDR-Examiner-Contacted.html

TODO:

- [ ] Add a `ExaminerContactedBoolean` field that can be used to set and get the value, which can generate the CodeableConcept for the user without them having to do it manually.

### DIFF
--- a/VRDR/DeathRecord.md
+++ b/VRDR/DeathRecord.md
@@ -348,7 +348,6 @@
   - [Get_Race_NHOPI_Literals()](#M-VRDR-IJEMortality-Get_Race_NHOPI_Literals 'VRDR.IJEMortality.Get_Race_NHOPI_Literals')
   - [Get_Race_OTHER_Literals()](#M-VRDR-IJEMortality-Get_Race_OTHER_Literals 'VRDR.IJEMortality.Get_Race_OTHER_Literals')
   - [Get_Race_W_Literals()](#M-VRDR-IJEMortality-Get_Race_W_Literals 'VRDR.IJEMortality.Get_Race_W_Literals')
-  - [Get_YNU()](#M-VRDR-IJEMortality-Get_YNU-System-String- 'VRDR.IJEMortality.Get_YNU(System.String)')
   - [HispanicOrigin()](#M-VRDR-IJEMortality-HispanicOrigin 'VRDR.IJEMortality.HispanicOrigin')
   - [HispanicOriginOther()](#M-VRDR-IJEMortality-HispanicOriginOther 'VRDR.IJEMortality.HispanicOriginOther')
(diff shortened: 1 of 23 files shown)


VRDR/DeathRecord.md
```diff
@@ -348,15 +348,13 @@
   - [Get_Race_NHOPI_Literals()](#M-VRDR-IJEMortality-Get_Race_NHOPI_Literals 'VRDR.IJEMortality.Get_Race_NHOPI_Literals')
   - [Get_Race_OTHER_Literals()](#M-VRDR-IJEMortality-Get_Race_OTHER_Literals 'VRDR.IJEMortality.Get_Race_OTHER_Literals')
   - [Get_Race_W_Literals()](#M-VRDR-IJEMortality-Get_Race_W_Literals 'VRDR.IJEMortality.Get_Race_W_Literals')
-  - [Get_YNU()](#M-VRDR-IJEMortality-Get_YNU-System-String- 'VRDR.IJEMortality.Get_YNU(System.String)')
```
rbclark, line 351: Is this file autogenerated? I manually removed these references since the functions are removed I just wanted to make sure I shouldn't be fully regenerating the documentation in some way.

hadleynet:
I wonder if the API should expose a simple boolean even if the underlying FHIR representation has changed to a codeable concept? Seems like an undesirable amount of work for a dev to have to create a codeable concept when using the API and, if they really wanted to work at that level, they could go grab the underlying Bundle via the GetBundle() method and work at that level instead.

rbclark:
> I wonder if the API should expose a simple boolean even if the underlying FHIR representation has changed to a codeable concept? Seems like an undesirable amount of work for a dev to have to create a codeable concept when using the API and, if they really wanted to work at that level, they could go grab the underlying Bundle via the GetBundle() method and work at that level instead.

If we want to keep it simple we should also considering changing it for `MilitaryService`, `AutopsyPerformed`, and `AutopsyResultsAvailable` since these are all also simple booleans that require a CodableConcept.

hadleynet:
We should be careful about API changes that could break existing users code. I don't know how many users we currently have but perhaps we should be adding parallel methods so ExaminerContacted, MilitaryService etc can be get/set as either boolean or codedConcept.

rbclark:
This now should work fine, however it has the slightly undesirable effect of causing Canary to display as follows:

Screen Shot 2020-06-29 at 5 24 50 PM

This is due to Examiner Contacted still technically being a boolean. I think the best long-term fix is to merge this for now and introduce the breaking change during the next major version. (We still technically haven't released a final 3.1.0 so we could technically get away with fully migrating this over now before we release and just labeling it as one of the major changes between the two versions).

PR #124, opened 2020-09-11
Bug fixes for RC4

hadleynet:
1. Ignore document reference identifiers that have no value.
2. Default death certificate number to 000000.

### DIFF
--- a/VRDR/DeathRecord.cs
+++ b/VRDR/DeathRecord.cs
@@ -193,9 +193,6 @@ namespace VRDR
             // Start with an empty certification.
             DeathCertification = new Procedure();
             DeathCertification.Id = Guid.NewGuid().ToString();
-            Identifier certificationIdentifier = new Identifier();
-            certificationIdentifier.Value = "000000";
-            DeathCertification.Identifier.Add(certificationIdentifier);
             DeathCertification.Subject = new ResourceReference("urn:uuid:" + Decedent.Id);
             DeathCertification.Meta = new Meta();
             string[] deathcertification_profile = { "http://hl7.org/fhir/us/vrdr/StructureDefinition/VRDR-Death-Certification" };
@@ -435,11 +432,8 @@ namespace VRDR
         {
             get
             {
-                if (DeathCertification != null && DeathCertification.Identifier != null && DeathCertification.Identifier.Count > 0)
-                {
-                    return DeathCertification.Identifier[0].Value;
-                }
-                return null; 
+                Identifier id = DeathCertification?.Identifier?.FirstOrDefault(i => i.Value != null && i.Value.Length > 0);
+                return id == null ? null : id.Value;
             }
             set
             {
--- a/VRDR/DeathRecord.cs
+++ b/VRDR/DeathRecord.cs
@@ -194,7 +194,7 @@ namespace VRDR
             DeathCertification = new Procedure();
             DeathCertification.Id = Guid.NewGuid().ToString();
             Identifier certificationIdentifier = new Identifier();
-            certificationIdentifier.Value = "";
+            certificationIdentifier.Value = "000000";
             DeathCertification.Identifier.Add(certificationIdentifier);
             DeathCertification.Subject = new ResourceReference("urn:uuid:" + Decedent.Id);
             DeathCertification.Meta = new Meta();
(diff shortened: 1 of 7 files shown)


VRDR/DeathRecord.cs
```diff
@@ -194,7 +194,7 @@ public DeathRecord()
             DeathCertification = new Procedure();
             DeathCertification.Id = Guid.NewGuid().ToString();
             Identifier certificationIdentifier = new Identifier();
-            certificationIdentifier.Value = "";
+            certificationIdentifier.Value = "000000";
```
pkmitre, line 197: In Canary, in the Connectathon record tests, if a user provides a record that has no certificate number Canary will report

```
Expected Certificate number: 00051
Found value: 000000
```
and

```
JSON

Expected:

{"identifier": "[
  {
    "value": "000051"
  }
]"}

How Canary interpreted your input:

{"identifier": "[
  {
    "value": "000000"
  }
]"}
```

Seems like this would be confusing...

VRDR/DeathRecord.cs
```diff
@@ -519,11 +519,9 @@ public string StateLocalIdentifier
         {
             get
             {
-                if (StateDocumentReference != null && StateDocumentReference.Identifier != null && StateDocumentReference.Identifier.Count() > 0)
-                {
-                    return StateDocumentReference.Identifier[0].Value;
-                }
-                return null;
+                // return first non-null/empty identifier value or null if non found
```
pkmitre, line 522: Tiny typo: `non` -> `none` or perhaps `not`

VRDR/DeathRecord.cs
hadleynet: Why does Canary not use exactly what the tester submitted rather than ingesting then recreating it (with possible default values like this)?

VRDR/DeathRecord.cs
hadleynet: Not a typo, non is a valid prefix meaning not.

VRDR/DeathRecord.cs
holmesie: On further consideration, I think the entire Identifier should be left out. The things in the constructor are bare minimums for referencing, and this Identifier isn't necessary for that.

hadleynet: The profile requires 1..* identifier on that element, if we leave it out then our default would be to generate an invalid entry.

pkmitre: Which brings us back to the approach I suggested of adding the identifier but not including the value at all, which is valid. Any reasons not to do that?

VRDR/DeathRecord.cs
pkmitre: I was referring to the second non:

> return first non-null/empty identifier value or null if **non** found

VRDR/DeathRecord.cs
hadleynet: An empty identifier value is invalid.

VRDR/DeathRecord.cs
hadleynet: Ah, in that case I agree, it should be none. Will fix.

VRDR/DeathRecord.cs
holmesie: Are we promising that just doing a `new DeathRecord()`, without including anything else, will return something that is "validator valid"? Is that the reason for including this in the first place?

hadleynet: Good question. At a minimum I think we'd want it to be valid FHIR but we probably don't want to promise compatibility with the VRDR or US Core profiles since that would require us to include a lot of default data. I'm leaning toward not including the identifier at all (valid FHIR, not valid against against VRDR) vs including an empty identifier (causes parser errors).

pkmitre: By "not including the value at all", I don't mean including a blank string, I mean not having a value field present in the identifier. AFAICT that's valid.

hadleynet: `identifier: [ {} ]` causes a parser error, identifier: [] is ok.

pkmitre: Is that what we get if we simply remove the line of code

```
certificationIdentifier.Value = <something>;
```

?

pkmitre: We'd also need to see how Canary handles that...

VRDR/IJEMortality.cs
```diff
@@ -1039,11 +1038,11 @@ public string AGE
         {
             get
             {
-                if (record.AgeAtDeath != null)
+                if (record.AgeAtDeath != null && !String.IsNullOrWhiteSpace(record.AgeAtDeath["value"]))
                 {
                     return record.AgeAtDeath["value"];
```
hadleynet, line 1043: I think you need to pad with leading zeroes

VRDR/IJEMortality.cs
```diff
@@ -1039,11 +1038,11 @@ public string AGE
...
                 }
-                return "";
+                return "999";
```
hadleynet, line 1045: I think you need to return 999 if AGETYPE is 9 regardless of the value of record.AgeAtDeath.

PR #126, opened 2021-01-26
Add additional fields to CodedResponseMessage

rbclark:
Added fields: CoderStatus, ShipmentNumber, NCHSReceiptDate, MannerOfDeath, IntentionalReject, ACMESystemRejectCodes, and PlaceOfInjury

Added a special case for RACEBRG which provides a different code when dealing with this special code.

General question for all of these fields: Do any of them need an accompanying code list? If so, where should such a code list come from?

### DIFF
--- a/VRDR.Messaging/CodingResponseMessage.cs
+++ b/VRDR.Messaging/CodingResponseMessage.cs
@@ -69,7 +69,7 @@ namespace VRDR
                 var list = new List<Tuple<string, Base>>();
                 foreach (KeyValuePair<HispanicOrigin, string> item in value)
                 {
-                    var part = Tuple.Create(item.Key.ToString(), 
+                    var part = Tuple.Create(item.Key.ToString(),
                         (Base)(new Coding("https://www.cdc.gov/nchs/data/dvs/HispanicCodeTitles.pdf", item.Value)));
                     list.Add(part);
                 }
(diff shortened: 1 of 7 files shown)


VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -140,14 +140,267 @@ public enum RaceCode
...
+                Record.Remove("int_rej");
+                if (value != null)
+                {
+                    if (value != 9 || value > 5 || value < 1) {
+                        throw new ArgumentException("Intentional Reject must equal 1-5, 9, or null.");
+                    }
+                    Record.Add("int_rej", new UnsignedInt((int)value));
+                }
+            }
+        }
```
rbclark, line 312: Question: should some of these be of the form

```json
{
    "name": "int_rej",
    "valueCoding": {
        "system": "<some code system>",
        "code": "<1-5, 9>"
    }
}
```

instead of

```json
{
    "name": "int_rej",
    "valueUnsignedInt": "<1-5, 9>"
}
```

VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -140,14 +140,267 @@ public enum RaceCode
...
+            /// <summary>Farm</summary>
+            Farm,
+            /// <summary>Other Specified Place</summary>
+            OtherSpecifiedPlace,
+            /// <summary>Unspecified Place</summary>
+            UnspecifiedPlace
+        }
+
+        /// <summary>Place of Injury (or null)</summary>
+        public PlaceOfInjuryEnum? PlaceOfInjury
```
rbclark, line 379: Question: Does this need to be a `Dictionary<PlaceOfInjuryEnum?, string>` in order to properly support values such as `OtherSpecifiedPlace`?

VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -140,14 +140,267 @@ public enum RaceCode
...
+                    var part = Tuple.Create(item.Key.ToString(),
+                        (Base)(new Coding(codeList, item.Value)));
                     list.Add(part);
                 }
                 Record.Add("race", list);
             }
         }
 
+        /// <summary>Coder Status (Numeric, 0-9)</summary>
+        public uint? CoderStatus
```
hadleynet, line 157: I wonder if this should be a code instead of an integer? That way you could include a code system that references the list of statuses ?

VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -140,14 +140,267 @@ public enum RaceCode
...
+            {
+                Record.Remove("ship");
+                if (value != null)
+                {
+                    Record.Add("ship", new FhirString(value));
+                }
+            }
+        }
+
+        /// <summary>NCHS Receipt Date Month (Numeric, 01-12 or null)</summary>
```
hadleynet, line 193: May want to check with users to see if a uint is going to require them to do string padding (e.g. 1 -> "01") for their needs. May want to switch to a string or offer a convenience method if so. Same for NCHSReceiptDay and (maybe for consistency) year.

VRDR.Messaging/CodingResponseMessage.cs
```diff
@@ -140,14 +140,267 @@ public enum RaceCode
...
+                Record.Remove("manner");
+                if(value != null)
+                {
+                    Record.Add("manner", new FhirString(value.ToString()));
+                }
+            }
+        }
+
+        /// <summary>Intentional reject (1-5, 9 or null)</summary>
+        public uint? IntentionalReject
```
hadleynet, line 295: Same comment re using a code string instead of a uint.

VRDR.Messaging/CodingResponseMessage.cs
hadleynet: I think so, see previous comments.

VRDR.Messaging/CodingResponseMessage.cs
hadleynet: I would probably go with two properties, one for the enum and a second optional one for the "specified place".

hadleynet (commented):
Comments inline.

VRDR.Messaging/CodingResponseMessage.cs
rbclark: I've gone ahead and changed this to be a Code, one question I have is should I still have logic to check if it is a valid value or should I rely on the user to properly implement the code system?

hadleynet: The VRDR library leaves it up to the user to ensure codes are valid. I'd recommend doing the same for VRDR.Messaging.

pkmitre (approved):

PR #129, opened 2021-04-22
Reorder death record properties to reflect order in standard death cert

smacadam:
This PR addresses [this ticket](https://boards.mitre.org/b/oaY5CRMEqJK3SsLrP/cdc-nvss/hu6PRoMNHZ45EH5sL)
Changed the priority values for death record properties to reflect the structure of the [US standard death certificate](https://www.cdc.gov/nchs/data/dvs/DEATH11-03final-ACC.pdf) 

To test
- Pull the updates and publish
- In the canary project, update the canary.csproj file to use the updated local vrdr-dotnet code
- Run canary and check the producing FHIR records and consuming FHIR records tools to see the updated field order and compare to the US standard death certificate

### DIFF
--- a/VRDR/DeathRecord.cs
+++ b/VRDR/DeathRecord.cs
@@ -426,7 +426,7 @@ namespace VRDR
         /// <para>// Getter:</para>
         /// <para>Console.WriteLine($"Death Certificate Number: {ExampleDeathRecord.Identifier}");</para>
         /// </example>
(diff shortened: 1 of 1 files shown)


rbclark (changes requested):
It seems like additionally reordering the sections themselves would make this reflect the standard death record format better. Do you have any thoughts on that @pkmitre?

rbclark (approved):

PR #138, opened 2021-06-16
Fhir ije race ethnicity bug

smacadam:
This is for ticket [NVSS-96](https://tracker.codev.mitre.org/browse/NVSS-96)

Initially, we thought this was a bug, but the test files were just formatted incorrectly (race and ethnicity were swapped). See ticket for the original test files. The updated test files are attached. I also made a small change to the logic for Hispanic Origin other. The previous logic made hispanic origin other and ije specific origins like Puerto Rican mutually exclusive.

**Testing**
To test run the following on the two test files
```
dotnet run 2ije round-trip-works.json > round-trip-works.ije
dotnet run ije2json round-trip-works.ije > round-trip-works-v2.json
```
The results should match the original json within the constraints of ije.

I also added two test cases in DeathRecord_Should.cs so we can verify our logic works in the future. To test run the following in VRDR.Tests
```
dotnet test
```
All tests should pass.

[testfiles.zip](https://github.com/nightingaleproject/vrdr-dotnet/files/6664413/testfiles.zip)


VRDR/IJEMortality.cs
```diff
@@ -1703,7 +1703,8 @@ public string DETHNIC4
                     return "N";
                 }
                 // We need to handle cases where hispanic origin is other with or without write-in
-                else if (hispanicOrigin != null && hispanicOrigin.Length > 0 && DETHNIC1 == "N" && DETHNIC2 == "N" && DETHNIC3 == "N")
```
rbclark, line 1706: The code seems functionally fine, for readability it may be a bit more readable if it were changed to the following pattern we use in a few other places.

```
string[] validHispanicOrigins = { "Cuban", "2182-4", "Puerto Rican", "2180-8", "Mexican", "2148-5" };

else if (hispanicOrigin != null && hispanicOrigin.Any(element => validHispanicOrigins.Contains(element)))
```

rbclark (approved):

PR #141, opened 2021-06-24
Fix nyc alternate debug

saulakravitz:
Fixed parsing of DeathLocationJurisdictionID.  Added test.
Added Valid method to IJEMortality that throws an exception if DSTATE is invalid. Invoke this at the end of KJEMortality(String).


VRDR/IJEMortality.cs
```diff
@@ -116,6 +126,18 @@ public DeathRecord ToDeathRecord()
             return this.record;
         }
 
+       /// <summary>Validates this IJE string.  Throws an exception if it is invalid, otherwise returns true</summary>
+        public Boolean Valid()
+        {
+            string dstate = this.DSTATE;
+            Boolean valid = dataLookup.JurisdictionNameToJurisdictionCode(dstate) != null;
+            string code = dataLookup.JurisdictionNameToJurisdictionCode(dstate);
+            Console.WriteLine("Valid: DSTATE=" + dstate + " code:" + code + " valid:" + valid );
```
holmesie, line 135: Left over debug print statement?

VRDR/MortalityData.cs
```diff
@@ -250,7 +262,68 @@ public string RaceCodeToRaceName(string code)
                 t => LinqHelper.EqualsInsensitive(t.Key, key)
             ).Value;
         }
-
+        // JurisdictionCodes uses IJE-defined two-character string as key, and provides the code defined in the US Vital Records Jurisdictions (NCHS) Value Set 
+        // all codes are from FIPS-5-2 except for YC which is from U.S. Board on Geographic Names (USGS - GNIS).   This is handled in the one use of this Dictionary in the code below.
+         public Dictionary<string, string> JurisdictionCodes = new Dictionary<string, string>
```
holmesie, line 267: This comment needs to follow the proper C# style to be picked up by the xml doc generator; take a look at the linting output you see likely locally, or the GitHub auto warnings right below this (I see it saying: `Missing XML comment for publicly visible type or member 'MortalityData.JurisdictionCodes'`).

VRDR/DeathRecord.cs
```diff
@@ -6228,13 +6167,10 @@ public string DeathLocationJurisdiction
...
                         CodeableConcept cc = (CodeableConcept)jurisdiction.Value;
-                        return cc.Coding[0].Display;
-                    }
-                    else
-                    {
-                        return DeathLocationLoc.Address.State ;  // This is a hack.   The Jurisdiction should be mandatory, but it isn't.  If it is absent, revert to using the state.  This will fail for NYC addresses.
+                        return MortalityData.JurisdictionCodeToJurisdictionName(cc.Coding[0].Code);
                     }
                 }
+                Console.Error.WriteLine("DeathLocationJurisdiction NOT found get: NULL");
```
holmesie, line 6173: Left over debug print statement?

VRDR/IJEMortality.cs
```diff
@@ -116,6 +126,18 @@ public DeathRecord ToDeathRecord()
             return this.record;
         }
 
+       /// <summary>Validates this IJE string.  Throws an exception if it is invalid, otherwise returns true</summary>
+        public Boolean Valid()
```
holmesie, line 130: This is very nit-picky, but did you consider alternative names for this method? Perhaps `Validate()` or something like that to give the impression that this will _perform_ the validation when you actually call it (and if there are problems it will raise an exception, rather than return false which I think the name implies it might do).

holmesie: Might also make more sense as a `private` method instead.

VRDR/IJEMortality.cs
```diff
@@ -81,8 +81,16 @@ public IJEMortality(string ije)
                 // Grab the field value
                 string field = ije.Substring(info.Location - 1, info.Length);
                 // Set the value on this IJEMortality (and the embedded record)
```
holmesie, line 83: I apparently can't create a comment on a line higher than this, so I'm going to put it here. In regards to the "let's make validation optional", here is my proposal:

Add a second optional parameter to this constructor, say:
```
public IJEMortality(string ije, bool validate = true)
```
and wrap line 93 in an if, like:
```
if (validate) {
  this.Valid();
}
```

That way you can force this constructor to not validate if you provide the optional parameter as true, otherwise it will always run.

holmesie (changes requested):
Looks great; left some comments

VRDR.CLI/Program.cs
```diff
@@ -563,12 +563,20 @@ static int Main(string[] args)
                 Console.WriteLine("Converting FHIR to IJE...\n");
                 DeathRecord d = new DeathRecord(File.ReadAllText(args[1]));
                 //Console.WriteLine(XDocument.Parse(d.ToXML()).ToString() + "\n");
-                IJEMortality ije1 = new IJEMortality(d);
-                //Console.WriteLine(ije1.ToString() + "\n\n");
-                IJEMortality ije2 = new IJEMortality(ije1.ToString());
+                IJEMortality ije1, ije2, ije3;
+                try {
+                ije1 = new IJEMortality(d);
```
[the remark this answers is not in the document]
saulakravitz: fixed

pkmitre (approved):

holmesie (approved):

PR #145, opened 2021-08-04
Handle case where ethnicity is empty in FHIR and Unknown in IJE, fix …

smacadam:
This PR addresses the issue Velan brought up on Zulip [Zulip](https://chat.fhir.org/#:~:text=previous%20way%20of-,validation,-(IJE%20vs%20FHIR)  and ticket [NVSS-195](https://tracker.codev.mitre.org/browse/NVSS-195)
If ethnicity data was missing in FHIR, the IJE conversion interpreted the missing data as Not Hispanic or Latino rather than unknown. 
Fix: If there is NO ethnicity data in FHIR, then all DETHNIC fields are set to U in IJE. And vice versa, if all DETHNIC fields are U in IJE, there will be no ethnicity data in FHIR. Because of the differences in how ethnicity is captured, individual DETHNIC fields set to "U" will not be preserved in a roundtrip, only if ALL the DETHNIC fields are U.

Some roundtrip IJE examples for the new expected behavior:
UUUU -> UUUU
NNHN -> NNHN
UUNH -> NNNH (unlikely scenario)
UUNU -> NNNN (unlikely scenario)

See test cases I added for more examples.

I brought this issue up with AMS in an IJE meeting. We can use this approach for now and in a future ballot explore using a missing data reason here or another solution.


saulakravitz (approved):
Look good.

Just need to clean up:
VRDR/IJEMortality.cs#L2674

XML comment is not placed on a valid language element

holmesie (approved):

PR #146, opened 2021-08-06
some annotations of the CodeSystems based on review of IG

saulakravitz:
Added additional constants and comments regarding codesystems.


smacadam (approved):
Looks good - one questions- are some of these value sets specifically for the new IG like data absent reason i.e. not used in the code yet? or were they just missed on the first iteration when you defined the CodeSystems?

saulakravitz:
> Looks good - one questions- are some of these value sets specifically for the new IG like data absent reason i.e. not used in the code yet? or were they just missed on the first iteration when you defined the CodeSystems?

Some aren't used yet, some were missed on my first pass.

PR #147, opened 2021-08-06
Cause-of-Death Condition and Condition-Contributing-To-Death -- add category field

saulakravitz:
Added category field to each of these.
Along the way rewrite the Condition-Contributing by factoring out the common 'constructor' code.   For each of the A-J (12) conditions, removed about 10 lines from three methods, for a total reduction of line count of about 360 LOC.   Very productive afternoon.


VRDR/DeathRecord.cs
```diff
@@ -58,6 +58,30 @@ public class DeathRecord
         /// <summary>Condition Contributing to Death.</summary>
         private Condition ConditionContributingToDeath;
 
+        /// <summary>Create a Cause Of Death Condition </summary>
+        private Condition CauseOfDeathCondition(int index){
+                    Condition CodCondition;
+                    Console.Error.WriteLine("CauseOfDeathCondition index = " + index);
```
smacadam, line 64: Should this log be removed? Or changed to an info level log

smacadam (changes requested):
Just needs the Console.Error log removed

smacadam:
Do you know if there are any tests for these fields?

VRDR/DeathRecord.cs
saulakravitz: Doh

PR #148, opened 2021-08-11
Use DPLACE for Death Location Type in roundtrip

smacadam:
The IJE field DPLACE was not implemented in the vrdr-dotnet library. Added in logic and tests to handle a roundtrip from IJE (DPLACE) to FHIR json (DeathLocationType)


VRDR/CodeSystems.cs
```diff
@@ -22,6 +22,8 @@ public static class CodeSystems
         public static string PH_YesNo_HL7_2x = "urn:oid:2.16.840.1.113883.12.136";
         /// <summary>PHINVADS Place of Occurrence.</summary>
         public static string PH_PlaceOfOccurrence_ICD_10_WHO = "urn:oid:2.16.840.1.114222.4.5.320";
+        /// <summary>PHINVADS Place Of Death.</summary>
+        public static string PH_PlaceOfDeath = "urn:oid:2.16.840.1.114222.4.11.7216"; // Should this use http://snomed.info/sct ?
```
saulakravitz, line 26: The 7216 OID is a valueset -- a collection of codes from one or more codesystems.  https://phinvads.cdc.gov/vads/ViewValueSet.action?oid=2.16.840.1.114222.4.11.7216 

In our code we are referencing code/codesystem/display triples.  So the code we need is the codesystem.
The above VS references PHINVADs' name for SNOMED, which is incredibly stupid, but that is what our code should cite.

So, SADLY, you should define a CodeSystem PH_SNOMED-CT (see https://phinvads.cdc.gov/vads/ViewCodeSystem.action?id=2.16.840.1.113883.6.96) with OID urn:oid2.16.840.1.113883.6.96

VRDR/IJEMortality.cs
```diff
@@ -1351,6 +1351,85 @@ public string MARITAL_BYPASS
...
+            }
+            set
+            {
+                if (!String.IsNullOrWhiteSpace(value))
+                {
+                    switch (value)
+                    {
+                        case "1":
+                            Dictionary_Set("DPLACE", "DeathLocationType", "code", "16983000");
+                            Dictionary_Set("DPLACE", "DeathLocationType", "system", CodeSystems.SCT);
```
saulakravitz, line 1390: All of the references to CodeSystems.SCT should be to CodeSystems.PH_SNOMED_CT

saulakravitz (changes requested):
See my comments.

PR #151, opened 2021-08-18
Birthdate date part absent

smacadam:
Addresses Tickets 

- [207](https://tracker.codev.mitre.org/browse/NVSS-207) STU2: Added Extension - Date Part Absent Reason to VRDR Decedent.birthdate
- [210](https://tracker.codev.mitre.org/browse/NVSS-210) STU2: Added Observation.dataAbsentReason to Resource Profile: Birth Record Identifier

Added the DateOfBirthDatePartAbsent field to the DeathRecord to capture the date part absent data in the _birthdate extension. If the extension is present, then the library expects all date parts to be in the extension.

Added the DataAbsentReason Observation to the BirthRecordId to capture the data absent reason for Birth Record Id.
 
Added tests to check the setters and getters for each field and created roundtrip tests to verify the translation from IJE to FHIR.

See [example](http://hl7.org/fhir/us/vrdr/2021Sep/Patient-51b806c8-566f-463e-8783-9fbf6be8161d.json.html) for an example of the json representation of birthdate with the new extension


VRDR/DeathRecord.cs
```diff
@@ -2956,6 +2956,92 @@ public string DateOfBirth
...
+        /// <para>// Setter:</para>
+        /// <para>ExampleDeathRecord.DateOfBirthDatePartReason = "1940-02-19";</para>
+        /// <para>// Getter:</para>
+        /// <para>Console.WriteLine($"Decedent Date of Birth Date Part Reason: {ExampleDeathRecord.DateOfBirthDatePartAbsent}");</para>
+        /// </example>
+        [Property("Date Of Birth Date Part Absent", Property.Types.TupleArr, "Decedent Demographics", "Decedent's Date of Birth Date Part.", true, "http://build.fhir.org/ig/HL7/vrdr/StructureDefinition-VRDR-Decedent.html", true, 14)]
+        [FHIRPath("Bundle.entry.resource.where($this is Patient)._birthDate.extension.where(url='http://hl7.org/fhir/us/vrdr/StructureDefinition/VRDR-Partial-date-part-absent-reason')", "_birthDate")]
+        public Tuple<string,string>[] DateOfBirthDatePartAbsent
+        {
+            get
```
saulakravitz, line 2971: Doesn't the date-part-absent extension get used in multiple places?
Could the code be factored so that the handling is in a method that is invoked by each place where the extension can be used?

VRDR/DeathRecord.cs
```diff
@@ -4034,13 +4120,17 @@ public string BirthRecordId
...
                     Bundle.AddResourceEntry(BirthRecordIdentifier, "urn:uuid:" + BirthRecordIdentifier.Id);
                 }
+                if (!String.IsNullOrWhiteSpace(value))
+                {
+                    BirthRecordIdentifier.Value = new FhirString(value);    
+                }
                 else
                 {
-                    BirthRecordIdentifier.Value = new FhirString(value);
+                    // TODO need to check for an existing data absent reason in the fhir
```
saulakravitz, line 4132: As per the comment, I think you need to check for the existence (or simply delete all existing ones) before creating a new one.

VRDR/DeathRecord.cs
```diff
@@ -2956,6 +2956,92 @@ public string DateOfBirth
...
+                        Extension yearAbsentPart = datePartAbsent.Extension.Where(ext => ext.Url == "year-absent-reason").FirstOrDefault();
+                        Extension monthAbsentPart = datePartAbsent.Extension.Where(ext => ext.Url == "month-absent-reason").FirstOrDefault();
+                        Extension dayAbsentPart = datePartAbsent.Extension.Where(ext => ext.Url == "day-absent-reason").FirstOrDefault();
+                        Extension yearPart = datePartAbsent.Extension.Where(ext => ext.Url == "date-year").FirstOrDefault();
+                        Extension monthPart = datePartAbsent.Extension.Where(ext => ext.Url == "date-month").FirstOrDefault();
+                        Extension dayPart = datePartAbsent.Extension.Where(ext => ext.Url == "date-day").FirstOrDefault();
+                        // Year part
+                        if (yearAbsentPart != null)
+                        {
+                            dateParts.Add(Tuple.Create("year-absent-reason", yearAbsentPart.Value.ToString())); // TODO change this to the code system value?
```
saulakravitz, line 2987: Ultimately these will be coded values, right?   Where do you do that conversion?

VRDR/IJEMortality.cs
```diff
@@ -286,6 +286,27 @@ private string DateTime_Get(string ijeFieldName, string dateTimeType, string fhi
             }
         }
 
+        /// <summary>Get the date part value from DateOfBirthDatePartAbsent if it's populated</summary>
+        private string BirthDate_Part_Absent_Get(string ijeFieldName, string dateType, string dateAbsentType)
```
saulakravitz, line 290: I don't understand this (due to my ignorance), so we need a review by someone who does ;-)

saulakravitz (commented):
Isn't the date-part-absent used in multiple places?
Can the handling in DeathRecord.cs be factored out?

VRDR/DeathRecord.cs
smacadam: Do you mean the conversion to IJE? The BirthDate_Part_Absent_Get function in IJEMortality checks for a date absent part and will return a string of "9"s the length of the given IJE field if there is an absent reason.

VRDR/DeathRecord.cs
smacadam: yes, I'll put it in a helper function for reuse

VRDR/DeathRecord.cs
smacadam: I looked into this more and it will be fine as is. The only time we need to use that is when we are converting between IJE and FHIR and need to create the data absent field. If we're just reading in fhir and using fhir then the data will be there - I ran a couple CLI json2json and checkJson tests to verify. I'll remove the todo comment

PR #162, opened 2021-08-26
AgeAtDeathAbsent andBirthRecordIDAbsent

saulakravitz:


smacadam:
A json2json test should preserve the data absent reason, but right now it just sets the value to "unknown". I tested with the DeathRecordBirthRecordDataAbsent.json fixture, the expected value is ask-unknown. My initial thought - either the value is overwritten by the default "unknown" somewhere, or the data absent value path isn't correct for parsing the json bundle.

smacadam (approved):

PR #167, opened 2021-09-10
Catch error on invalid jurisdiction id

smacadam:
This addresses the issue Veronique reported via email, captured in [ticket 254](https://tracker.codev.mitre.org/browse/NVSS-254). 
I had a call with Veronique to discuss the issue. She had message that was formatted with an old version of the library where the jurisdiction id extension looked like this 
```
        "extension": [
          {
            "url": "http://hl7.org/fhir/us/vrdr/StructureDefinition/Location-Jurisdiction-Id",
            "valueCodeableConcept": {
              "text": "MD"
            }
          }
        ],
```
instead of this
```
       "extension": [
          {
            "url": "http://hl7.org/fhir/us/vrdr/StructureDefinition/Location-Jurisdiction-Id",
            "valueCodeableConcept": {
              "coding": [
                {
                  "system": "urn:oid:2.16.840.1.113883.6.92",
                  "code": "25",
                  "display": "MA"
                }
              ],
              "text": "MA"
            }
          }
        ],
```
this caused the library to throw an index out of bounds error. To handle the case I added in another check in DeathRecord.cs. Now, instead of an error, the invalid extension will just return null and put an empty string when converting to IJE. Tests were added to validate the new code. 

Reviewer please verify: is this how we want the library to handle malformed/outdated extensions?

### DIFF
--- a/VRDR.Tests/DeathRecord.Tests.csproj
+++ b/VRDR.Tests/DeathRecord.Tests.csproj
@@ -34,6 +34,7 @@
     <None Update="fixtures/json/EmptyMessage.json" CopyToOutputDirectory="PreserveNewest" />
     <None Update="fixtures/json/EmptySubmission.json" CopyToOutputDirectory="PreserveNewest" />
     <None Update="fixtures/json/ExtractionErrorMessage.json" CopyToOutputDirectory="PreserveNewest" />
+    <None Update="fixtures/json/InvalidJurisdictionId.json" CopyToOutputDirectory="PreserveNewest" />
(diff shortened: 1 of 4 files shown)


pkmitre:
This seems reasonable, but I thought I remember Veronique suggesting that the library should instead just raise a different error that, when sent back to the jurisdiction, makes it clearer what the problem is. Does that make sense, especially since this is a required field? Or, will having the blank in IJE cause an error to be sent back through some other mechanism?

PR #169, opened 2021-09-13
Update README and the field description for jurisdiction id

smacadam:
Changes for Connectathon Sep 2021

- For ticket 251, Update README.md
  - Updated the library README to link to the STU2 IG
  - Fixed the formatting for the links in the Library Usage section
  - Updated the pregnancy status example in the README
- For ticket 258, Canary doesn't allow user to enter codeable concept for jurisdiction id
  - This was intentional, the user should enter the jurisdiction name and the library will do the conversion
  - The field description was updated so the user knows what value they should enter, see screen shot below
Screen Shot 2021-09-13 at 12 07 56 PM

### DIFF
--- a/VRDR/DeathRecord.cs
+++ b/VRDR/DeathRecord.cs
@@ -5880,7 +5880,7 @@ namespace VRDR
         /// <para>// Getter:</para>
         /// <para>Console.WriteLine($"Death Location Jurisdiction: {ExampleDeathRecord.DeathLocationJurisdiction}");</para>
         /// </example>
-        [Property("Death Location Jurisdiction", Property.Types.String, "Death Investigation", "The Jurisdiction Name, the library will generate the codeable concept.", true, locationJurisdictionExtPath, false, 16)]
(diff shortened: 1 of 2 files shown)


VRDR/DeathRecord.cs
```diff
@@ -5880,7 +5880,7 @@ public string DateOfDeathPronouncement
         /// <para>// Getter:</para>
         /// <para>Console.WriteLine($"Death Location Jurisdiction: {ExampleDeathRecord.DeathLocationJurisdiction}");</para>
         /// </example>
-        [Property("Death Location Jurisdiction", Property.Types.String, "Death Investigation", "Vital Records Jurisdiction of Death Location.", true, locationJurisdictionExtPath, false, 16)]
+        [Property("Death Location Jurisdiction", Property.Types.String, "Death Investigation", "The Jurisdiction Name, the library will generate the codeable concept.", true, locationJurisdictionExtPath, false, 16)]
```
pkmitre, line 5883: I think we want to still make it clear that this is the location of death; I suggest

```Vital Records Jurisdiction of Death Location (two character jurisdiction code, e.g. CA)```

pkmitre (changes requested):

pkmitre (approved):

PR #190, opened 2021-12-23
Incorporate auto-generated code into VRDR.Messaging/.  add CertifierRole VS to ruby script and to ValueSets.cs / 

saulakravitz:
This PR incorporates:
  1) minor change of adding the CertifierRole VS to ValueSets.cs for Pete's Nightingale effort
  2) major change to VRDR.Messaging that incorporates accessors generated directly from the Vital Records Messaging IG into the VRDR.Messaging library.   If the IG changes, it is a minute to regenerate the accessors.   Bringing the test cases into line remains a PITA.   Some of the test cases have been updated.   
  
  #2 requires serious review.
  
  Remaining decisions:
     1) encoding of 'regurgitated fields' in the Messaging IG:  currently they are encoded as per IJE.   Paul will decide.
     2) naming of fields.   All new fields are named as per IJE.   Older fields are named close-to IJE, but sometimes differ in case (cert_no instead of CERT_NO) or more major differences (underlying_cause_of_death instead of MAN_UC).
     
 Some of the tests fail because the test cases fail to comply with the current definitions for record and entity-axis cause o death.   Those would be cleaned up before merging this PR (obviously).
 
 the changes are still worth a review before moving forward.


VRDR.Messaging/ParameterUtils.cs
```diff
@@ -0,0 +1,358 @@
...
+            Parameters.ParameterComponent f = Record.GetSingle(field);
+            if (f == null){
+                return null;
+            } else {
+                FhirString fs = (FhirString)f.Value;
+                string s = (string)fs.Value;
+                return s;
+            }
+        }
+        // These should be simply referenced from VRDR/Program.cs....Not sure why they aren't visible from there.
```
smacadam, line 294: The DictToCoding function in DeathRecord.cs is private, try making it public to use it here. It seems like more of a util function so I think its appropriate to make it public

VRDR.Messaging/ParameterUtils.cs
```diff
@@ -0,0 +1,358 @@
...
+                    coding.Display = dict["display"];
+                }
+            }
+            return coding;
+        }
+
+        /// <summary>Convert a FHIR Coding to a "code" Dictionary</summary>
+        /// <param name="coding">a FHIR Coding.</param>
+        /// <returns>the corresponding Dictionary representation of the code.</returns>
+        private Dictionary<string, string> CodingToDict(Coding coding)
```
smacadam, line 319: same for this function

VRDR.Messaging/AliasMessage.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <returns></returns>
+        internal AliasMessage(Bundle messageBundle) : base(messageBundle)
+        {
+        }
+
+        /// <summary>Constructor that takes a VRDR.DeathRecord and creates a message to Alias that record.</summary>
+        /// <param name="record">the VRDR.DeathRecord to create a AliasMessage for.</param>
+        public AliasMessage(DeathRecord record) : this()
+        {
+            ExtractBusinessIdentifiers(record);
```
smacadam, line 29: this function will need to be updated to use the new field names

VRDR.Messaging/Parameters.cs
```diff
@@ -0,0 +1,1277 @@
+
```
smacadam, line 1: Is this the file of generated code? If it is, it should have a comment saying it's generated and not to edit because changes will be overwritten

VRDR.Messaging/AliasMessage.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using Hl7.Fhir.Model;
+
+namespace VRDR
+{
+    /// <summary>Class <c>AliasMessage</c> indicates that Alias information is being provided for a previously submitted DeathRecordSubmission message.
+    ///    Modeled on void message.
+    /// </summary>
+    public class AliasMessage : BaseMessage
```
smacadam, line 9: are all the alias fields part of the generated code or base message and that's why there's no fields in this class?

VRDR.Messaging/Parameters.cs
rbclark: To add to this, it may be useful as part of the CI to automate the generation of this file in order to provide something to check against in CI to make sure nobody has manually modified the content. The advantage being that would restrict anyone from accidentally modifying the content of these files and then having their changes lost down the road.

VRDR.Messaging/VRDRMessaging.xml
```diff
@@ -29,10 +29,32 @@
             <value>the message id.</value>
         </member>
         <member name="P:VRDR.AckMessage.BlockCount">
+            DELETE THIS   vvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
```
rbclark, line 32: I'm rather confident that this file is autogenerated, so there should be no need to do anything special here if the actual code is updated.

rbclark:
General comment: It seems logical to do a full sweep and update all fields to be named per the IJE. Having some match and some not seems like it will be somewhat confusing. I'd imaging such a change will only get harder to make over time.

smacadam:
Parts related to VRDR are for master and Messaging are for IG

mitre-jim:
Pulled out the non-messaging changes into pull request https://github.com/nightingaleproject/vrdr-dotnet/pull/197. Still need to sort out the VRDR.Messaging changes to get merged into the new IG branch.

mitre-jim:
Non-IG changes already merged to master. Messaging IG changes incorporated into #207

PR #200, opened 2022-03-09
Move the client library functions into vrdr-dotnet, update README with examples

smacadam:
The client authentication, token handling, post requests, and get requests have been moved into the vrdr-dotnet library. This will make it easier for jurisdictions to reuse code without having to read through the entire client reference implementation.

### DIFF
--- a/VRDR.Client/Client.cs
+++ b/VRDR.Client/Client.cs
@@ -1,0 +1,104 @@
+﻿using System;
+using System.Text;
+using System.Threading.Tasks;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using Newtonsoft.Json.Linq;
+using VRDR;
+using RestSharp;
+
+namespace nvssclient.lib;
+public class Client
+{
+    /// <summary>The API url</summary>
+    public String Url { get; }
+    /// <summary>Whether the client is running locally</summary>
+    public bool LocalTesting { get; }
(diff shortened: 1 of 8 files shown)


mitre-jim:
Nullable warnings are enabled for the project, leading to several compiler warnings. Most of this is due to the empty default constructors for the two classes not setting values. Do we _need_ these constructors? Or is it fair to say that these two objects (Credentials, Client) are only valid when all of the values are provide (i.e. second constructor)? The sample code only uses the constructors with arguments. Unless there is use case for partially constructing these objects and setting those properties later, I'd suggest removing the no-arg constructors.

VRDR.Client/Client.cs
```diff
@@ -0,0 +1,147 @@
...
+public class Client
+{
+    /// <summary>The API url</summary>
+    public String Url { get; }
+    /// <summary>Whether the client is running locally</summary>
+    public bool LocalTesting { get; }
+    /// <summary>The credentials to access the API server</summary>
+    public Credentials Credentials { get; }
+    /// <summary>The token to access the API server</summary>
+    public string Token { get; set; }
```
mitre-jim, line 21: Token gets checked for null/empty a few times, and its value gets used for the authentication header, but its value never gets set anywhere. How is it supposed to get a value assigned?

pkmitre (approved):

PR #202, opened 2022-04-05
Nightingale conversion fixes and additional updates

pkmitre:
This pull request makes a handful of not-necessarily-related changes:

1. Updates and adds to the code for converting from the Nightingale data format to FHIR
2. Adds the CertificationRoleHelper method
3. Documents the new helper methods in the README
4. Copies the Connectathon record generation code into VRDR from Canary so that it can be accessed from the VRDR.CLI; once this is merged the version of the file in Canary can be removed and Canary can refer to this version
5. Adds functionality to the VRDR.CLI application for generating Connectathon records from the command line
6. Adds a ruby script for generating a large quantity of FHIR records using the Connectathon records
7. Adds a ruby script for converting an Excel file with records in pseudo-IJE format into a set of FHIR records for testing
8. Removes support for the now-obsolete netcoreapp2.1

### DIFF
--- a/VRDR.HTTP/Nightingale.cs
+++ b/VRDR.HTTP/Nightingale.cs
@@ -30,15 +30,15 @@ namespace VRDR.HTTP
             SetYesNoValueDictionary(values, "didTobaccoUseContributeToDeath.didTobaccoUseContributeToDeath", record, "TobaccoUse");
             SetYesNoValueDictionary(values, "meOrCoronerContacted.meOrCoronerContacted", record, "ExaminerContacted");
 
-            if (GetValueDict(record.CertificationRole, "code") == "434651000124107")
+            if (record.CertificationRoleHelper == "434651000124107")
             {
                 values["certifierType.certifierType"] = "Physician (Certifier)";
             }
-            else if (GetValueDict(record.CertificationRole, "code") == "434641000124105")
+            else if (record.CertificationRoleHelper == "434641000124105")
             {
                 values["certifierType.certifierType"] = "Physician (Pronouncer and Certifier)";
             }
-            else if (GetValueDict(record.CertificationRole, "code") == "455381000124109")
+            else if (record.CertificationRoleHelper == "455381000124109")
             {
                 values["certifierType.certifierType"] = "Medical Examiner";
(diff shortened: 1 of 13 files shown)


VRDR.HTTP/Nightingale.cs
```diff
@@ -610,54 +647,78 @@ private static void SetStringValueDeathRecordDictionary(DeathRecord deathRecord,
             prop.SetValue(deathRecord, dict, null);
         }
 
-        /// <summary>Set a Yes/No (coded) value on a DeathRecord.</summary>
-        private static void SetYesNoValueDeathRecordCode(DeathRecord deathRecord, string property, string value)
+        /// <summary>Set a Yes/No (coded) value on a DeathRecord using Boolean helpers.</summary>
+        private static void SetYesNoValueDeathRecordBoolean(DeathRecord deathRecord, string property, string value)
```
smacadam, line 651: It seems like this function is useful for other parts of the library, should it be in VRDR vs VRDR.HTTP?

VRDR.HTTP/Nightingale.cs
```diff
@@ -610,54 +647,78 @@ private static void SetStringValueDeathRecordDictionary(DeathRecord deathRecord,
...
+            }
+            else
+            {
+                prop.SetValue(deathRecord, null);
             }
         }
 
         /// <summary>Set a Yes/No (coded) value on a DeathRecord.</summary>
-        private static void SetYesNoValueDeathRecordBool(DeathRecord deathRecord, string property, string value)
+        private static void SetYesNoValueDeathRecordCode(DeathRecord deathRecord, string property, string value)
```
smacadam, line 674: Same as question above

smacadam (commented):
Questions on where the Nightingale helper functions should live within the library

VRDR.HTTP/Nightingale.cs
pkmitre: Seems reasonable, but should only be moved when it's actually first used outside of VRDR.HTTP (YAGNI principle)

smacadam (approved):

PR #219, opened 2022-04-15
351 decedent profile and input race ethnicity

smacadam:
Changes for the decedent profile and the input race and ethnicity profile. This includes changes for every fields in the two profiles EXCEPT the DOB fields which will be addressed at the end of the update with all other partial date fields.


saulakravitz (approved):
Didn't review every line, but checked overall approach.
Looks good.  
Why haven't tests completed?  Are they stuck?

VRDR/DeathRecord.cs
```diff
@@ -282,6 +301,19 @@ public DeathRecord()
             string[] decedent_profile = { "http://hl7.org/fhir/us/vrdr/StructureDefinition/VRDR-Decedent" };
             Decedent.Meta.Profile = decedent_profile;
 
+            // Start with an empty race and ethinicity observation
```
mitre-jim, line 304: Why doesn't this use the `CreateRaceEthnicityObs` method? Looks like it duplicates the code.

smacadam: The CreateRaceEthnicityObs calls "AddReferenceToComposition" which throws a null reference exception so the function doesn't work here

PR #221, opened 2022-04-16
Decedent age

mitre-jim:
Updates the AgeAtDeath to align with the IG; includes addition of age at death edit bypass flag.


VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -2115,11 +2115,53 @@ public void AgeAtDeath_RoundTrip()
...
+            Assert.Equal("http://hl7.org/fhir/us/vrdr/CodeSystem/vrdr-bypass-edit-flag-cs", SetterDeathRecord.AgeAtDeathEditBypassFlag["system"]);
+            Assert.Equal("0", SetterDeathRecord.AgeAtDeathEditBypassFlag["code"]);
+            Assert.Equal("Edit Passed", SetterDeathRecord.AgeAtDeathEditBypassFlag["display"]);
+
+            flag = new Dictionary<string, string>();
+            flag.Add("system", "http://hl7.org/fhir/us/vrdr/CodeSystem/vrdr-bypass-edit-flag-cs");
+            flag.Add("code", "1");
+            flag.Add("display", "Edit Failed, Data Queried, and Verified");
+            SetterDeathRecord.AgeAtDeathEditBypassFlag = flag;
+            Assert.Equal("http://hl7.org/fhir/us/vrdr/CodeSystem/vrdr-bypass-edit-flag-cs", SetterDeathRecord.AgeAtDeathEditBypassFlag["system"]);
```
saulakravitz, line 2140: All Profile and Extension references in CS code should use ProfileURL and ExtensionURL constants.

VRDR/DeathRecord.cs
```diff
@@ -6348,22 +6348,57 @@ public string DeathLocationTypeHelper
...
                     Quantity quantity = (Quantity)AgeAtDeathObs.Value;
-                    quantity.Unit = GetValue(value, "unit");
+                    quantity.Unit = extractedValue;
                     AgeAtDeathObs.Value = quantity;
+                    AgeAtDeathDataAbsentBoolean = false; // if age is present, cancel the data absent reason
                 }
               }
         }
 
+        [Property("Age At Death Data Absent Reason (Code)", Property.Types.Dictionary, "Death Investigation", "Age At Death Edit Bypass Flag.", true, "http://build.fhir.org/ig/HL7/vrdr/StructureDefinition-vrdr-decedent-age.html", true, 2)]
```
saulakravitz, line 6370: The reference should use IGURL.DecedentAge.  (All references to profile html pages should use this pattern).

VRDR/DeathRecord.cs
```diff
@@ -6348,22 +6348,57 @@ public string DeathLocationTypeHelper
...
+        [PropertyParam("code", "The code used to describe this concept.")]
```
saulakravitz, line 6371: Missing XML comment for this method.

VRDR/DeathRecord.cs
```diff
@@ -6348,22 +6348,57 @@ public string DeathLocationTypeHelper
...
+        [PropertyParam("system", "The relevant code system.")]
+        [PropertyParam("display", "The human readable version of this code.")]
+        [PropertyParam("text", "Additional descriptive text.")]
+        [FHIRPath("Bundle.entry.resource.where($this is Observation).where(code.coding.code='39016-1').extension.where(url='http://hl7.org/fhir/us/vrdr/ValueSet/vrdr-edit-bypass-01-vs')", "")]
+        public Dictionary<string, string> AgeAtDeathEditBypassFlag
```
saulakravitz, line 6376: The extension URL should be ExtensionURL.BypassEditFlag.   The URL you have is the URL of the value set used by this instance of the extension.

saulakravitz (commented):
Need to add test content in json and xml files for AgeAtDeathBypass.
Ditto for round trip tests.

VRDR.Tests/DeathRecord_Should.cs
mitre-jim: There is not currently a constant defined in URLs.cs for this particular URL. Is this the incorrect URL and I need to use a different one, or is this correct and I need to add the URL (and would it fall under "profile", "extension", or "IG" urls)?

saulakravitz: ExtensionURL.BypassEditFlag -- 'id' of the extension
IGURL.BypassEditFlag -- 'url of documentation for the extension'.   not sure this is needed anywhere for extensions.

mitre-jim: It was defined in CodeSystem.cs instead of URLs.cs

VRDR/IJEMortality.cs
```diff
@@ -1211,11 +1167,24 @@ public string AGE_BYPASS
         {
```
saulakravitz, line 1167: I think the EditBypass Flag handling would be simpler if you converted the Boolean helper in DeathRecord.cs to a helper that took a code value (string).  Like PregnancyStatusEditFlagHelper.   Allows you to use the mapping infrastructure instead of writing code.

saulakravitz (approved):
See my comments about the Bypass.
I think it should follow the same pattern as PregnancyStatusEditFlagHelper which involves getting rid of the Boolean helper in DeathRecord.cs and replacing it so you can use the mapping tools.

mitre-jim:
Rebased and forced a push to resolve conflicts, but my local IG-develop was behind, so the conflict did not get resolved. Will pull and repeat

PR #222, opened 2022-04-16
Removed unused RestSharp dependency.

mitre-jim:
Remove the `using` statement and pulled the reference from the project file.


saulakravitz (approved):
I have no idea what this does, but looks innocent enough ;-)

PR #227, opened 2022-04-21
Location updates

saulakravitz:
Most of the needed updates to Locations (Death, Injury) and Funeral Home.
Current status is that one of the Mortality tests fails, but I can't figure out why.
My eternal gratitude to the person who figures out why DSTATE is " ".   Everything else in 'dotnet test' runs successfully, although the tests need to be built out a bit more to cover additional fields.


mitre-jim:
It looks like the state (and other parts of the address) gets set correctly initially, but setting CITYCODE_D overwrites/erases the existing address fields.

Possible cause is copy/paste error? The setter refers to "COUNTYTEXT_D" rather than "CITYCODE_D". Might not be the _cause_, but probably needs to be fixed?

Inside `Dictionary_Geo_Set`, it gets the Dictionary from the DeathRecord in order to update the field. However, it first checks whether the dictionary exists and the value is not already set. If the dictionary does not exist, or if the key is already set (in this case it is), it _overwrites the whole dictionary with a new one_, rather than updating/setting the field.

mitre-jim:
So `CITYTEXT_D` and `CITYCODE_D` both want to set the "addressCity" value in the `DeathLoactionAddress` dictionary. Whichever one gets set first will set the value initially. Then if either of them gets set again, they will overwrite it.

I think this bug is probably a bit bigger than this specific issue. Using `Dictionary_Geo_Set` to set a value for any field in a dictionary that has already been set will result in wiping the entire dictionary and replacing it with a new one with just that value set.

smacadam:
To add to Jim's finding, the regular Dictionary_Set function has the same logic. Unless there was a reason why the logic was implemented this way, this function should be fixed too and tests should be added to verify the fix.

smacadam:
Rebasing to resolve conflicts. If others need to make changes, delete your local version of the branch and pull from origin

smacadam:
Need to make sure the tests are re-enabled, get answer from standards team on the BSTATE

saulakravitz:
> I think this bug is probably a bit bigger than this specific issue. Using `Dictionary_Geo_Set` to set a value for any field in a dictionary

It it fixing these methods, or adding a new method?

smacadam:
> > Need to make sure the tests are re-enabled, get answer from standards team on the BSTATE
> 
> WHich tests? The BSTATE issue is a non-issue. Basically a redundancy introduced by NAPHSIS that NCHS has no interest in eliminating.

Re-enabled the CLI test to check all fields by removing our if statement for COUNTYTEXT_I and COUNTYTEXT_D since the issue is resolved with the new changes

smacadam (approved):

PR #229, opened 2022-04-22
353 death certificate document

smacadam:
Updates for the death certificate document http://build.fhir.org/ig/HL7/vrdr/StructureDefinition-vrdr-death-certificate-document.html 
Can someone confirm the StateAuxiliaryIdentifier1 is the correct state identifier to get in the ExtractBusinessIdentifiers function?

### DIFF
--- a/VRDR.CLI/Program.cs
+++ b/VRDR.CLI/Program.cs
@@ -41,7 +41,8 @@ namespace VRDR.CLI
                 deathRecord.CertificationRole = certificationRole;
 
                 // State Local Identifier
-                deathRecord.StateLocalIdentifier = "42";
+                Tuple<string, string>[] ids = new Tuple<string, string>[]{Tuple.Create(ExtensionURL.AuxiliaryStateIdentifier1, "000000000042")};
+                deathRecord.StateLocalIdentifier = ids;
 
                 // MannerOfDeathType
                 Dictionary<string, string> mannerOfDeathType = new Dictionary<string, string>();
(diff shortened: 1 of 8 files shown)


VRDR.CLI/Program.cs
```diff
@@ -41,7 +41,8 @@ static int Main(string[] args)
                 deathRecord.CertificationRole = certificationRole;
 
                 // State Local Identifier
-                deathRecord.StateLocalIdentifier = "42";
+                Tuple<string, string>[] ids = new Tuple<string, string>[]{Tuple.Create(ExtensionURL.AuxiliaryStateIdentifier1, "000000000042")};
```
saulakravitz, line 44: THis looks like a syntactic step backward.
Should be deathRecord.StateLocalIdentifier1 = 
                  deathRecordStateLocalIdentifier2 = 
                  
           The two identifiers are not an array, but rather two separate extensions.

VRDR.Messaging/BaseMessage.cs
```diff
@@ -102,7 +102,16 @@ protected void ExtractBusinessIdentifiers(DeathRecord from)
             {
                 this.CertificateNumber = certificateNumber;
             }
-            this.StateAuxiliaryIdentifier = from?.StateLocalIdentifier;
+            // take the first state local identifier if it exists
+            if (from?.StateLocalIdentifier != null && from.StateLocalIdentifier.Length > 0)
+            {
+                this.StateAuxiliaryIdentifier = from.StateLocalIdentifier[0].Item2;
```
saulakravitz, line 108: These shouldn't be an array.  They are separate variables.  Auxno can be empty when Auxno2 has a value.

saulakravitz (approved):
I think the syntax in the old examples is much easier to use.
The code in DeathRecord.cs needs to be updated to reference the precise extension URL for both AUXNo and AUXNO2.   Currently, it just looks at the first two extensions of identifier.

PR #236, opened 2022-05-02
Connectathon tests (3)

saulakravitz:
Generated three test cases from Alaina's spreadsheet, finding one error in the spreadsheet along the way.
Testcases are specified form the IJE perspective, which is the way that they are handed to the testing jurisdictions.


VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -2851,6 +2851,17 @@ public void Get_EmergingIssues()
             Assert.Equal("AAAAAAAAAAAAAAAAAAAA", ((DeathRecord)XMLRecords[0]).EmergingIssue20);
         }
 
+        [Fact]
+        public void CheckConnectathonRecord1()
+        {
+            DeathRecord dr1 = VRDR.Connectathon.FideliaAlsup();
+            IJEMortality ije = new IJEMortality(dr1, false); // Don't validate since we don't care about most fields
```
pkmitre, line 2858: Aren't the connectathon records intended to be complete records, and so shouldn't they validate? What validation failures show up?

VRDR/Connectathon.cs
```diff
@@ -42,809 +38,278 @@ record = JavierLuisPerez(partialRecord: true);
...
-                dladdress.Add("addressZip", " 78201");
-                dladdress.Add("addressCountry", "US");
-                record.DispositionLocationAddress = dladdress;
-                record.DispositionLocationName = "River Cemetary";
-            }
+    /// <summary>Generate the Twila Hilty example record</summary>
+    public static DeathRecord TwilaHilty()
+    {
+    	IJEMortality ije = new IJEMortality(""); // blank IJE
+		ije.DOD_YR = "2022";
```
pkmitre, line 45: Code should use consistent indentation.

tools/generate_connectathon_testcases.rb
```diff
@@ -0,0 +1,58 @@
+require "pry"
```
pkmitre, line 1: Scripts should have a header that explains 1) what it does and 2) how to use it. See e.g., generate_url_strings_from_VRDR_IG.rb

Also, it looks like this script generates some file fragments that you then need to assemble? Why not just spit out the full output file? Use a template like in generate_url_strings_from_VRDR_IG.rb

pkmitre (changes requested):

VRDR.Tests/DeathRecord_Should.cs
saulakravitz: I'm just trying to create a blank IJEMortality, and the constructor where you pass in a blank string doesn't work.  This is a workaround:  create an empty DeathRecord, and use the IJEMortality constructor that takes a DeathRecord to create a blank record.   Can't validate because DSTATE is blank, and that is a validation error.   I corrected the comment and made all three cases consistent wrt how the IJEMOrtality record is created.

tools/generate_connectathon_testcases.rb
saulakravitz: 1) added documentation
2) didn't take it to the level of generate_url_strings_from_VRDR_IG.rb....didn't think the ROI justified it.

pkmitre (approved):

PR #250, opened 2022-05-09
Messaging README and documentation updates

pkmitre:
This pull request:

1. Updates README.md to have a correct example of creating a coded response message
2. Updates doc/Messaging.md to have correct examples of creating and reading coded response messages
3. Updates the CLI code to have working code to parse and write out contents of coded response messages
4. Updates the Messaging tests to read and write source and destination
5. Updates the versions table in README.md to include Messaging IG information and IG links


saulakravitz (approved):
Looks good.  Messaging.md. still refers to NCHSIdentifier.
Also the full example is written from the perspective of IJE, and the partial example is written from the perspective of FHIR.  So, they kind of complement each other, rather than illuminating with detail.   

This is something we can refine.

saulakravitz:
Nice recyclying of PRogram.cs as an example for building a death record!
Looks good.

PR #251, opened 2022-05-10
eliminate CauseOfDeathPathway, add lineNumber component to Part1Cause…

saulakravitz:
…OfDeath

- Code and examples for COD1X (Part1) Interval was not aligned with IG.
- Swapped in new examples, and fixed codes to look for INTERVALx is a valueString, rather than the text field from a CodeableConcept.
- Cleanup some stray IGURL, Codesystem, and ProfileURL constants.
- Fix some FHIRPaths.
- Stripped the deathrecord constructor down to the bare minimum.  It needs Collection, and everything the collection refers to directly (not as entries) to exist, but not all of the random observations.
- Corrected DeathCertificate (Composition) sections for several elements.
- Delete a bunch of methods that aren't used (e.g., UsualOccupationCode).

### DIFF
--- a/VRDR/DeathRecord.cs
+++ b/VRDR/DeathRecord.cs
@@ -65,6 +65,7 @@ namespace VRDR
             Certifier.Meta = new Meta();
             string[] certifier_profile = { ProfileURL.Certifier  };
             Certifier.Meta.Profile = certifier_profile;
+            // Not linked to Composition or inserted in bundle, since this is run before the composition exists.
         }
         // ///<summary>The Mortician.</summary>
         //private Practitioner Mortician;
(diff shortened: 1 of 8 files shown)


VRDR/DeathRecord.cs
```diff
@@ -39,34 +39,51 @@ public class DeathRecord
...
         /// <summary>The Certification.</summary>
         private Procedure DeathCertification;
 
-        /// <summary>Create Death Certification.</summary>
-        private void CreateDeathCertification(){
-            CreateEmptyDeathCertification();
-            AddReferenceToComposition(DeathCertification.Id, "DeathCertification");
-            Bundle.AddResourceEntry(DeathCertification, "urn:uuid:" + DeathCertification.Id);
-        }
+        // /// <summary>Create Death Certification.</summary>
```
pkmitre, line 78: This looks like a commented out duplicate

pkmitre (approved):

PR #261, opened 2022-05-16
Defend against NPE

mitre-jim:


VRDR/DeathRecord.cs
```diff
@@ -1106,6 +1106,7 @@ public string CertifiedTime
             }
             set
             {
+                // TODO: Create new Composition if it is null (how should it be initialized?)
```
pkmitre, line 1109: I think we want to change these TODO's because they assume we want to create a new Composition rather than potentially raising an exception (at least in some cases)

pkmitre (changes requested):

pkmitre (approved):

PR #269, opened 2022-05-23
Handle name field being required for Location resources

pkmitre:
This pull request

1. Supports creation of VRDR records where there is no data for Location resource name fields where a non-blank name string is required using "BLANK" as a placeholder
2. Supports consumption of similar records
3. Incidentally cleans up tests a bit bit not reloading test files when already loaded

### DIFF
--- a/VRDR.Tests/DeathRecord_Should.cs
+++ b/VRDR.Tests/DeathRecord_Should.cs
@@ -2135,6 +2135,10 @@ namespace VRDR.Tests
         {
             SetterDeathRecord.DispositionLocationName = "Bedford Cemetery";
             Assert.Equal("Bedford Cemetery", SetterDeathRecord.DispositionLocationName);
+            SetterDeathRecord.DispositionLocationName = "";
+            Assert.Null(SetterDeathRecord.DispositionLocationName);
+            SetterDeathRecord.DispositionLocationName = null;
+            Assert.Null(SetterDeathRecord.DispositionLocationName);
         }
 
         [Fact]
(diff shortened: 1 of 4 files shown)


VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -2135,6 +2132,10 @@ public void Set_DispositionLocationName()
         {
             SetterDeathRecord.DispositionLocationName = "Bedford Cemetery";
             Assert.Equal("Bedford Cemetery", SetterDeathRecord.DispositionLocationName);
+            SetterDeathRecord.DispositionLocationName = "";
```
saulakravitz, line 2135: should be "BLANK" (or similar)...."" triggers a warning about empty strings.

VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -2135,6 +2132,10 @@ public void Set_DispositionLocationName()
...
+            Assert.Null(SetterDeathRecord.DispositionLocationName);
+            SetterDeathRecord.DispositionLocationName = null;
```
saulakravitz, line 2137: should be "BLANK" (or similar)...."" triggers a warning about empty strings.

VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -2135,6 +2132,10 @@ public void Set_DispositionLocationName()
...
+            Assert.Null(SetterDeathRecord.DispositionLocationName);
```
saulakravitz, line 2138: Not sure this is the right test.

VRDR.Tests/DeathRecord_Should.cs
pkmitre: Look at the code for the property, it handles this correctly putting BLANK in the FHIR representation.

VRDR.Tests/DeathRecord_Should.cs
pkmitre: Look at the code for the property, it handles this correctly putting BLANK in the FHIR representation.

saulakravitz (approved):

PR #274, opened 2022-05-26
Add parser to handle just a bundle and just a BaseMessage

smacadam:
Our BaseMessage Parse function returns a specific type of Message, ex CauseOfDeathCodingMessage. We ran into an issue where a TRX message created by Veronique's code was aligned with the new version of the library but the API had not been updated and threw an error when it couldn't parse the message. This change allows us to keep the parsing on the API server as generic as possible by only parsing the json to a BaseMessage. We need to still parse it as a BaseMessage so we can access the message id when we add the message to the response bundle.

This will allow more messages to get returned in the responses from the API. It will be the client's responsibility to verify the specific message type.

### DIFF
--- a/VRDR.Messaging/BaseMessage.cs
+++ b/VRDR.Messaging/BaseMessage.cs
@@ -465,7 +465,7 @@ namespace VRDR
         /// <param name="source">the XML or JSON serialization of a FHIR Bundle</param>
         /// <param name="permissive">if the parser should be permissive when parsing the given string</param>
         /// <returns>The deserialized bundle object</returns>
-        public static Bundle ParseGenericBundle(string source, bool permissive = false)
+        private static Bundle ParseGenericBundle(string source, bool permissive = false)
         {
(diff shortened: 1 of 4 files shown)


VRDR.Messaging/BaseMessage.cs
```diff
@@ -441,21 +441,44 @@ public string NCHSIdentifier
...
+            return message;
+        }
+
+        /// <summary>
+        /// Parse an XML or JSON serialization of a FHIR Bundle. 
+        /// </summary>
+        /// <param name="source">the XML or JSON serialization of a FHIR Bundle</param>
+        /// <param name="permissive">if the parser should be permissive when parsing the given string</param>
+        /// <returns>The deserialized bundle object</returns>
+        public static Bundle ParseGenericBundle(string source, bool permissive = false)
```
pkmitre, line 468: Should we make this private?

pkmitre:
We should probably add a new test parsing some arbitrary message (using one of the old trx messages might be a good bet) using the new functionality.

PR #275, opened 2022-06-07
Make sure a value is a valueString before parsing it as a string / Fix Documentation

saulakravitz:
1) Fix bug where a value that was expected to be a string is provided as a CodeableConcept, and parsed wrong.
2) Fix a lot of the documentation entries in DeathRecord.cs that were incorrect, or OBE
3) Regenerate the DeathRecord.md file


VRDR/DeathRecord.cs
```diff
@@ -6726,13 +6734,13 @@ public string PregnancyStatusEditFlagHelper
             }
         }
 
-        /// <summary>Examiner Contacted Boolean. This is a conenience method, to access the code use ExaminerContacted instead.</summary>
+        /// <summary>Examiner Contacted Helper. This is a conenience method, to access the code use ExaminerContacted instead.</summary>
```
pkmitre, line 6737: Might as well edit the "conenience" typo if we're changing this line.

VRDR/DeathRecord.csproj
```diff
@@ -13,9 +15,13 @@
   <ItemGroup>
     <PackageReference Include="Hl7.Fhir.R4" Version="1.6.0" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
-    <!-- <PackageReference Include="Vsxmd" Version="1.4.2">
+    <PackageReference Include="Vsxmd" Version="1.4.5">
```
pkmitre, line 18: I don't think we want a dependency on Vsxmd to actually ship (we don't want users of our package to pull in this dependency), so I'd suggest 1) commenting it out by default, and 2) including a note on that in the section you added on generating the file

pkmitre (changes requested):

VRDR/DeathRecord.cs
```diff
@@ -9919,7 +9935,7 @@ public string IntentionalRejectHelper
         /// </value>
         /// <example>
         /// <para>// Setter:</para>
-        /// <para>ExampleDeathRecord.AcmeSystemReject = 3;</para>
+        /// <para>ExampleDeathRecord.AcmeSystemReject = "3";</para>
```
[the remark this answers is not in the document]
saulakravitz: Actually, this is the documentation for the AcmeSystemRejectHelper....so fixed that.

PR #278, opened 2022-06-15
Remove second addition of injury location object to output bundle

reece-berens:
Adding the injury location object to the death record bundle already occurs in the CreateInjuryLocationLoc method. Removing the code to add that object to the bundle in the InjuryLocationAddress setter will only add the injury location object to the output bundle once, which is what is expected.


saulakravitz (approved):
Yes, this fixes a refactoring error, where the AddReference and Bundle.AddResourceEntry are included in the CreateInjuryLocationLoc() method.

PR #282, opened 2022-06-27
Fix four problems that came up in Testing Event

saulakravitz:
fix 
1) VRDR.CLI/1.MOR which had an invalid value for SpouseALive of Y (should be…1,2,8,9).   
2) handling of field with blanks in TRANSPORT and CERTL...should be a noop.
3) COnsistently refer to AlaskanNative. (had some ALaskanNative and some AlaskaNative)....also fixed IG.
4) valueTime should be hh:mm:ss, not hh:mm according to https://build.fhir.org/datatypes.html#time


VRDR/DeathRecord.cs
```diff
@@ -1382,7 +1382,11 @@ public string CertificationRoleHelper
             }
             set
             {
-                if ((value != null) && !VRDR.Mappings.CertifierTypes.FHIRToIJE.ContainsKey(value))
+                if (value == null || String.IsNullOrEmpty(value.Trim())){
```
mitre-jim, line 1385: Can use `String.IsNullOrWhitespace` instead of combining `Trim` and `IsNullOrEmpty`

saulakravitz: modified as per suggestion...thx

VRDR/DeathRecord.cs
```diff
@@ -7402,11 +7406,15 @@ public string TransportationRoleHelper
             }
             set
             {
+                if (value == null || String.IsNullOrEmpty(value.Trim())){
```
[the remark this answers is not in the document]
saulakravitz: modified as per suggestion...thx

mitre-jim (approved):

mitre-jim:
There are several occurrences of AlaskaNative that did not get updated:

- VRDR/RaceAndEthnicity.cs (the summary comment) and the associated generated XML doc (VRDR/DeathRecord.xml)
- VRDR/Tests/fixtures/json/BlankLocationNames.json
- Several VRDR.CLI files (two occurrences each):
   - 1.json
   - 1.xml
   - 1submit.json
   - 1_wJurisdiction.json
   - 1_wJurisdiction.xml

PR #285, opened 2022-07-08
clean up unused MortalityData.cs stuff

saulakravitz:
- Deleted a lot of unused code in MortalityData.cs
- Commented out one line in VRDR.HTTP/Nightingale.cs that is probably OBE.


smacadam:
Errors occurring in Nightingale where it references deleted functions

VRDR/MortalityData.cs
```diff
@@ -324,1008 +172,15 @@ public string RaceCodeToRaceName(string code)
...
             {"UT","Utah"},
             {"VT","Vermont"},
-            {"VI","Virgin Islands"},
             {"VA","Virginia"},
             {"WA","Washington"},
             {"WV","West Virginia"},
             {"WI","Wisconsin"},
             {"WY","Wyoming"},
-            {"YC","New York City"},
-            {"TT","Test Jurisdiction"} // This should only be used for testing with NCHS.
```
pkmitre, line 336: Don't we need TT for testing?

VRDR/IJEMortality.cs
```diff
@@ -655,18 +655,18 @@ private void Dictionary_Geo_Set(string ijeFieldName, string fhirFieldName, strin
...
-                        dictionary.TryGetValue(keyPrefix + "County", out county);
-                        if (!String.IsNullOrWhiteSpace(state) && !String.IsNullOrWhiteSpace(county))
-                        {
-                            string city = dataLookup.StateNameAndCountyNameAndPlaceCodeToPlaceName(state, county, value);
-                            if (!String.IsNullOrWhiteSpace(city))
-                            {
-                                dictionary[key] = city;
-                            }
-                        }
+                        // string state = null;
```
pkmitre, line 658: Why is some code deleted and some other code just commented out? I propose that if we comment out some code that we want to leave around in a commented out state that there should always be some comment prose that describes 1) why the code is no longer needed and 2) why we might need it in the future such that commenting it out is the right decision.

pkmitre (changes requested):
A pull request should never have a merge in it, it should just be a series of linear commits

VRDR/MortalityData.cs
saulakravitz: It is there....just a github rendering thing.

VRDR/IJEMortality.cs
saulakravitz: Code should be deleted...  Deleted it.

saulakravitz:
> 

The original pull request was on July 8th...needed to merge in changes in the intervening month before creating the PR, no?

pkmitre (approved):

PR #293, opened 2022-08-10
Trx and mre validation tools in CLI

saulakravitz:
Added tools in the VRDR.CLI to convert from Coding Content Bundles to/from MRE and TRX formats, as well as comparing JSON and TRX/MRE formatted data via translation to IJE.
Added two tests to .github/workflows/run-tests.yml.... test that FHIR->TRX and FHIR-MRE appear to work.

### DIFF
--- a/VRDR.CLI/Program.cs
+++ b/VRDR.CLI/Program.cs
@@ -61,9 +61,10 @@ namespace VRDR.CLI
 ";
         static int Main(string[] args)
         {
-            if ((args.Length == 0) || ((args.Length == 1) && (args[0] == "help"))){
+            if ((args.Length == 0) || ((args.Length == 1) && (args[0] == "help")))
+            {
                 Console.WriteLine(commands);
-                return(0);
+                return (0);
             }
             else if (args.Length == 1 && args[0] == "fakerecord")
             {
@@ -482,7 +483,7 @@ namespace VRDR.CLI
                     // Grab the field value
                     string field = Convert.ToString(property.GetValue(ije1, null));
                     // Print the key/value pair to console
-                    Console.WriteLine( info.Name + ": " + field);
+                    Console.WriteLine(info.Name + ": " + field);
                 }
 
                 return 0;
@@ -538,7 +539,7 @@ namespace VRDR.CLI
                 DeathRecord indr = new DeathRecord(File.ReadAllText(args[1]));
                 DeathRecord outdr = new DeathRecord();
                 List<PropertyInfo> properties = typeof(DeathRecord).GetProperties().ToList();
-                foreach(PropertyInfo property in properties)
+                foreach (PropertyInfo property in properties)
                 {
                     if (property.Name.Contains("CausesOfDeath") || property.Name.Contains("CertifierQualification"))
                     {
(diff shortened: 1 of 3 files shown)


VRDR/DeathRecord.cs
```diff
@@ -389,13 +389,15 @@ private void CreateDeathDateObs()
             DeathDateObs.Meta.Profile = deathdate_profile;
             DeathDateObs.Status = ObservationStatus.Final;
             DeathDateObs.Code = new CodeableConcept(CodeSystems.LOINC, "81956-5", "Date+time of death", null);
-            DeathDateObs.Subject = new ResourceReference("urn:uuid:" + Decedent.Id);
+            DeathDateObs.Subject = new ResourceReference("urn:uuid:" + (Decedent!=null?Decedent.Id:"000000000"));
```
pkmitre, line 392: I think the right thing to do here is to not set the Subject if there isn't a Subject present.

VRDR.CLI/Program.cs
```diff
@@ -910,5 +986,132 @@ private static string Truncate(string value, int length)
                 return value.Substring(0, length);
             }
         }
+ private static IJEMortality mre2ije(string mrefilename){
```
pkmitre, line 989: Can you get back in the habit of code formatting to use our coding style for new code (brackets are always on a line on their own)?

VRDR/DeathRecord.cs
```diff
@@ -389,13 +389,15 @@ private void CreateDeathDateObs()
...
             // A DeathDate can be represented either using the PartialDateTime or the valueDateTime; we always prefer
             // the PartialDateTime representation (though we'll correctly read records using valueDateTime) and so we
             // by default set up all the PartialDate extensions with a default state of "data absent"
             DeathDateObs.Value = new FhirDateTime();
             DeathDateObs.Value.Extension.Add(NewBlankPartialDateTimeExtension(true));
-            AddReferenceToComposition(DeathDateObs.Id, "DeathInvestigation");
+            if(Decedent!=null){
```
pkmitre, line 398: Checking the Decedent to determine whether to add this reference is pretty opaque; once you make the change to not set the DeathDateObs.Subject if there isn't a Decedent just check to see if the DeathDateObs has a Subject set.

pkmitre, line 398: Another code formatting reminder (brackets).

VRDR/DeathRecord.cs
```diff
@@ -492,7 +494,9 @@ private void CreateDeathLocation()
             DeathLocationLoc.Meta.Profile = deathlocation_profile;
             DeathLocationLoc.Type.Add(new CodeableConcept(CodeSystems.LocationType, "death", "death location", null));
             DeathLocationLoc.Name = DeathRecord.BlankPlaceholder; // We cannot have a blank string, but the field is required to be present
-            AddReferenceToComposition(DeathLocationLoc.Id, "DeathInvestigation");
+            if(Decedent!=null){
```
pkmitre, line 497: Why is DeathLocationLoc being added to the Composition related to Decedent being present or not?

pkmitre (changes requested):

VRDR/DeathRecord.cs
saulakravitz: @pkmitre: Only DeathCertificateDocument includes a Composition and a Decedent.  So, it looks like I'm using a Decedent as a proxy for this.   Perhaps, AddReferenceToComposition should just return if there is no Composition?

pkmitre (approved):

PR #294, opened 2022-08-10
fix bug submitted by Reece(ID) via zulip 

saulakravitz:
Fix bug with injury incident value that was encoded as valueString instead of value.text [see](https://chat.fhir.org/#narrow/stream/179301-Death-on-FHIR/topic/Injury.20Incident.20in.20CoD.20Coding.20Bundle)

### DIFF
--- a/VRDR/DeathRecord.cs
+++ b/VRDR/DeathRecord.cs
@@ -7280,11 +7280,7 @@ namespace VRDR
                 CodeableConcept concept = InjuryIncidentObs?.Value as CodeableConcept;
                 if(concept != null)
                 {
-                    Dictionary<string, string> dict = CodeableConceptToDict(concept);
-                    if (dict.ContainsKey("text"))
-                    {
-                            return (dict["text"]);
-                    }
+                    return concept.Text;
                 }
                 return null;
             }
--- a/VRDR/DeathRecord.cs
+++ b/VRDR/DeathRecord.cs
@@ -7277,9 +7277,9 @@ namespace VRDR
         {
             get
             {
-                if (InjuryIncidentObs?.Value != null && InjuryIncidentObs.Value as CodeableConcept != null)
+                CodeableConcept concept = InjuryIncidentObs?.Value as CodeableConcept;
                 {
-                    Dictionary<string, string> dict = CodeableConceptToDict((CodeableConcept)InjuryIncidentObs.Value);
+                    Dictionary<string, string> dict = CodeableConceptToDict(concept);
                     if (dict.ContainsKey("text"))
                     {
                             return (dict["text"]);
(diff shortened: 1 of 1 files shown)


VRDR/DeathRecord.cs
```diff
@@ -7270,9 +7270,13 @@ public string InjuryDescription
         {
             get
             {
-                if (InjuryIncidentObs?.Value != null)
+                if (InjuryIncidentObs?.Value != null && InjuryIncidentObs.Value as CodeableConcept != null)
```
mitre-jim, line 7273: The `as` check makes the first null comparison redundant, so only one should be needed.
Perhaps instead of casting it twice (in the `if` with `as` and then again on line 7275), this should be something like

```c#
CodeableConcept concept = InjuryIncidentObs?.Value as CodeableConecpt;
if (concept != null)
{
...
}
```

saulakravitz: done

VRDR/DeathRecord.cs
```diff
@@ -7285,7 +7290,7 @@ public string InjuryDescription
                 {
                     CreateInjuryIncidentObs();
                 }
-                InjuryIncidentObs.Value = new FhirString(value);
+                InjuryIncidentObs.Value = new CodeableConcept(null, null, null, value);
```
mitre-jim, line 7293: Does this not need a system/code value (or display text)?

saulakravitz: Nope.  The key idea is that we are passing a string value in the value.text field, because it is entirely possible that somebody will want to use the coded content of the value.   For now, all NCHS wants is a text string.
So, the new CodeableConcept(null, null, null, value); is creating  value.text only, and leaving value.coding.code empty.

VRDR/DeathRecord.cs
```diff
@@ -7270,9 +7270,14 @@ public string InjuryDescription
         {
             get
             {
-                if (InjuryIncidentObs?.Value != null)
+                CodeableConcept concept = InjuryIncidentObs?.Value as CodeableConcept;
+                if(concept != null)
                 {
-                    return Convert.ToString(InjuryIncidentObs.Value);
+                    Dictionary<string, string> dict = CodeableConceptToDict(concept);
```
pkmitre, line 7276: Why not just return concept.Text instead of going through the trouble of converting it to a dictionary? The one difference I see is that converting causes null to be returned in the case of an empty string... doesn't seem an important distinction here?

pkmitre (changes requested):

VRDR/DeathRecord.cs
saulakravitz: Good simplification.

pkmitre (approved):

PR #302, opened 2022-08-22
Initial Split of VRDR/DeathRecord.cs into 6 smaller files

saulakravitz:
Initial Split of VRDR/DeathRecord.cs into 6 smaller files:  fields, constructors, input properties, output properties, createMethods and utils


VRDR/DeathRecord_ctors.cs
```diff
@@ -0,0 +1,665 @@
+using System;
```
smacadam, line 1: Is this file for constructors? I wasn't sure based on filename

pkmitre: Is there a letter shortage? A good coding practice to adopt is to spell things out. The trade off of typing a few extra characters is well worth minimizing years of potential confusion on the part of people trying to understand.

pkmitre: Does changing DeathRecord_codedProperties.cs to just DeathRecord_properties.cs work? What does the "coded" signify?

VRDR/DeathRecord_ctors.cs
```diff
@@ -0,0 +1,665 @@
...
+                    Extension replaceExt = new Extension(ExtensionURL.ReplaceStatus , DictToCodeableConcept(ReplaceStatus) );
+                    mortRosterBundle.Meta.Extension.Add(replaceExt);
+            }
+            Extension aliasExt = new Extension(ExtensionURL.AliasStatus, new FhirBoolean(alias));
+            mortRosterBundle.Meta.Extension.Add(aliasExt);
+            return mortRosterBundle;
+        }
+
+
+/////////////////////////////////////////////////////////////////////////////////
```
pkmitre, line 330: The indenting on this line seems to have gotten munged.

Maybe run a formatter on all the files, since this is a time when the noise in a diff won't impact anything?

VRDR/DeathRecord_fields.cs
```diff
@@ -0,0 +1,158 @@
+using System;
```
pkmitre, line 1: Maybe move these in with the create methods? Put each one with the create method that sets it up?

pkmitre: Reasoning: locality is often helpful in organizing code. Thinking through "adding a new field" you'd put the field and the builder in the same place.

pkmitre (changes requested):
I think I understand "...codedProperties.cs" as a file name now. What does the "input" of "...inputProperties.cs" signify?

I also suggest adding a comment at the top of every file outlining what goes in that file.

smacadam:
Change inputProperties to submissionProperties and codedProperties to responseProperties

VRDR/DeathRecord_fields.cs
saulakravitz: consolidated as per your recommendation.
It is an improvement over the status quo ante, since now the methods are systematically adjacent to their associated fields...previously they were proximate, but it was sloppy.

saulakravitz:
OK...now it is 5 files...

pkmitre (approved):

PR #306, opened 2022-08-31
workaround for missing vs unknown

saulakravitz:


VRDR/IJEMortality.cs
```diff
@@ -2992,7 +3005,14 @@ public string TOI_UNIT
...
             get
             {
-                return "M"; // Military time
+                if (record.InjuryIncidentTimeSet())
+                {
+                    return "M"; // Military time
+                }
+                else
+                {
+                    return ""; // Military time
```
smacadam, line 3014: should this be " " ?

smacadam (approved):

PR #310, opened 2022-09-14
support NCHS 5 character ICD-like codes (without period) on all coded…

saulakravitz:
… fields

All ICD10 fields except RAC are 5 characters (with 5th documented as 'reserved').
RAC uses 5th position for pregnancy indicator.
All fields except RAC are now 5-character capable.

Simplified the conversion between NCHS and official ICD10 formats.   The regular expressions are interesting as documentation, but irrelevant for the actual conversion which involves adding or deleting a decimal point.

### DIFF
--- a/VRDR/IJEMortality.cs
+++ b/VRDR/IJEMortality.cs
@@ -777,26 +777,28 @@ namespace VRDR
         /// <summary>NCHS ICD10 to actual ICD10 </summary>
         private string NCHSICD10toActualICD10(string nchsicd10code)
         {
-            Regex ICD10rgx = new Regex(@"^[A-Z]\d{2}(\.\d){0,1}$");  // ICD10 codes only have one character after the decimal
-            Regex NCHSICD10rgx = new Regex(@"^[A-Z]\d{2,4}$");       // NCHS tacks on an extra character to some ICD10 codes, e.g., K7210 (K27.10)
-            string code;
-            nchsicd10code = nchsicd10code.Trim();
-            if (ICD10rgx.IsMatch(nchsicd10code))
+            // ICD-10 diagnosis codes always begin with a letter (except U) followed by a digit.
+            // The third character is usually a digit, but could be an A or B [1].
+            // After the first three characters, there may be a decimal point, and up to three more alphanumeric characters.
+            // These alphanumeric characters are never U. Sometimes the decimal is left out.
+            // Regex ICD10regex = new Regex(@"^[A-TV-Z][0-9][0-9AB].?[0-9A-TV-Z]{0,4}$");
+            // NCHS ICD10 codes are the same as above for the first three characters.
+            // The decimal point is always dropped.
+            // Some codes have a fourth character that reflects an actual ICD10 code.
+            // NCHS tacks on an extra character to some ICD10 codes, e.g., K7210 (K27.10)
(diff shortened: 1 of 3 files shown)


VRDR/IJEMortality.cs
```diff
@@ -777,8 +777,8 @@ private void Set_MappingIJEToFHIR(Dictionary<string, string> mapping, string ije
         /// <summary>NCHS ICD10 to actual ICD10 </summary>
         private string NCHSICD10toActualICD10(string nchsicd10code)
         {
-            Regex ICD10rgx = new Regex(@"^[A-Z]\d{2}(\.\d){0,1}$");
-            Regex NCHSICD10rgx = new Regex(@"^[A-Z]\d{2,3}$");
+            Regex ICD10rgx = new Regex(@"^[A-Z]\d{2}(\.\d){0,1}$");  // ICD10 codes only have one character after the decimal
```
pkmitre, line 780: Can't ICD codes can have multiple characters after the decimal? Note https://www.webpt.com/blog/understanding-icd-10-code-structure/

This is mostly relevant for the comments, but might make sense to update the regex to match.

VRDR/IJEMortality.cs
```diff
@@ -777,8 +777,8 @@ private void Set_MappingIJEToFHIR(Dictionary<string, string> mapping, string ije
...
+            Regex NCHSICD10rgx = new Regex(@"^[A-Z]\d{2,4}$");       // NCHS tacks on an extra character to some ICD10 codes, e.g., K7210 (K27.10)
             string code;
             nchsicd10code = nchsicd10code.Trim();
             if (ICD10rgx.IsMatch(nchsicd10code))
```
pkmitre, line 784: Does this case (code is already an ICD10 code so don't change it) basically come into play for three character codes?

pkmitre (changes requested):

VRDR/IJEMortality.cs
saulakravitz: Looking at https://www.johndcook.com/blog/2019/05/05/regex_icd_codes/, 
The letter can't be U, so they recommend:
[A-TV-Z][0-9][0-9AB]\.?[0-9A-TV-Z]{0,4}

VRDR/IJEMortality.cs
saulakravitz: Yes.    A 3 Char NCHS code is a valid ICD10 code.

VRDR/IJEMortality.cs
```diff
@@ -777,27 +777,28 @@ private void Set_MappingIJEToFHIR(Dictionary<string, string> mapping, string ije
...
+            // The decimal point is always dropped.
+            // Some codes have a fourth character that reflects an actual ICD10 code.
+            // NCHS tacks on an extra character to some ICD10 codes, e.g., K7210 (K27.10)
+            // Regex NCHSICD10regex = new Regex(@"^[A-TV-Z][0-9][0-9AB][0-9A-TV-Z]{0,2}$");
+            string code = "";
+
+            if (!String.IsNullOrEmpty(nchsicd10code))
             {
-                code = nchsicd10code;
+                 code = nchsicd10code.Trim();
```
pkmitre, line 794: Indenting is off by 1 extra space... otherwise makes sense to me!

pkmitre (changes requested):

pkmitre (approved):

PR #312, opened 2022-09-16
link coding messages to the message they code just like acks

saulakravitz:


VRDR.Messaging/DemographicsCodingMessage.cs
```diff
@@ -70,6 +95,24 @@ protected override Bundle MessageBundleRecord
                 return deathRecord?.GetDemographicCodedContentBundle();
             }
         }
+        /// <summary>The id of the death record submission/update message that was coded to produce the content of this message</summary>
+        /// <value>the message id.</value>
+        public string SubmittedMessageId
```
pkmitre, line 100: Same comment as above re: "CodedMessageId"

pkmitre (changes requested):

pkmitre (approved):

PR #318, opened 2022-10-10
SSN length checking fix

RobertScalfani:
This PR updates the logic for SSN numbers to check that their length and formats are correct.

### DIFF
--- a/VRDR/IJEMortality.cs
+++ b/VRDR/IJEMortality.cs
@@ -1086,16 +1086,15 @@ namespace VRDR
                 string ssn = record.SSN;
                 if (!String.IsNullOrWhiteSpace(ssn))
                 {
-                    ssn = ssn.Replace("-", string.Empty).Replace(" ", string.Empty);
-                    if (ssn.Length != info.Length)
+                    string formattedSSN = ssn.Replace("-", string.Empty).Replace(" ", string.Empty);
+                    if (formattedSSN.Length != info.Length)
                     {
-                        validationErrors.Add($"Error: FHIR field {fhirFieldName} contains string '{ssn}' which is not the expected length for IJE field {ijeFieldName} of length {info.Length}");
+                        validationErrors.Add($"Error: FHIR field {fhirFieldName} contains string '{ssn}' which is not the expected length (without dashes or spaces) for IJE field {ijeFieldName} of length {info.Length}");
                     }
-                    return Truncate(ssn, info.Length).PadRight(info.Length, ' ');
+                    return Truncate(formattedSSN, info.Length).PadRight(info.Length, ' ');
                 }
                 else
                 {
-                    validationErrors.Add($"Error: FHIR field {fhirFieldName} is missing data for IJE Field {ijeFieldName}");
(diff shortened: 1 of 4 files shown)


VRDR/IJEMortality.cs
```diff
@@ -1077,21 +1077,46 @@ public string SEX_BYPASS
         [IJEField(15, 191, 9, "Social Security Number", "SSN", 1)]
         public string SSN
         {
+
             get
             {
+                string fhirFieldName = "SSN";
+                string ijeFieldName = "SSN";
+                IJEField info = FieldInfo(ijeFieldName);
```
pkmitre, line 1085: When we're dealing with only a single field (SSN in this case) we don't need to reference the property via FieldInfo, we can just access it directly (like you actually do with record.SSN) and use the numeric field lengths directly. We only need FieldInfo if we want to do this dynamically for an arbitrary string (like we do in the helper methods).

VRDR/IJEMortality.cs
```diff
@@ -1077,21 +1077,46 @@ public string SEX_BYPASS
...
+                    {
+                        validationErrors.Add($"Error: FHIR field {fhirFieldName} contains string '{ssn}' which is not the expected length (without dashes or spaces) for IJE field {ijeFieldName} of length {info.Length}");
+                    }
+                    return Truncate(formattedSSN, info.Length).PadRight(info.Length, ' ');
                 }
                 else
                 {
-                    return "";
+                    return new String(' ', info.Length);
                 }
```
pkmitre, line 1099: So, the above could just be

```
    if (formattedSSN.Length != 9)
    {
        validationErrors.Add($"Error: FHIR field SSN contains string '{ssn}' which is not the expected length (without dashes or spaces) for IJE field SSN of length 9");
    }
    return Truncate(formattedSSN, 9).PadRight(9, ' ');
}
else
{
    return new String(' ', 9);
}
```

VRDR/IJEMortality.cs
```diff
@@ -1077,21 +1077,46 @@ public string SEX_BYPASS
...
+                string fhirFieldName = "SSN";
+                string ijeFieldName = "SSN";
+                if (!String.IsNullOrWhiteSpace(value))
+                {
+                    string ssn = value.Trim();
+                    if (ssn.Contains("-") || ssn.Contains(" "))
+                    {
+                        validationErrors.Add($"Error: IJE field {ijeFieldName} contains string '{value}' which cannot contain ` ` or `-` characters for FHIR field {fhirFieldName}.");
+                    }
+                    IJEField info  = FieldInfo(ijeFieldName);
```
pkmitre, line 1112: Same thought here about using the field names and lengths...

VRDR/IJEMortality.cs
pkmitre: Could argue that 9 is somewhat of a magic number the way I'm suggesting... happy to discuss

pkmitre (changes requested):

pkmitre (approved):

PR #319, opened 2022-10-14
Multiple bug fixes

pkmitre:
This pull request

1. Updates the CLI to pull data from the correct offsets in MRE and TRX files
2. Updates the CLI to provide clearer output when comparing MRE and TRX files to their FHIR equivalents
3. Updates the CLI to support creating a void range message
4.  Fixes a null reference bug when reading records that have literal race fields present with no actual value
5. Updates the manual API testing tools to support status messages
6. Updates the manual API testing tools to support single threaded submission
7. Updates the manual API testing tools to improve error handling

### DIFF
--- a/tools/manual_api_testing/push.rb
+++ b/tools/manual_api_testing/push.rb
@@ -19,6 +19,16 @@ end
 
 credentials = YAML.load(File.read(CONFIG_PATH))
 
+if ARGV.include?('--single')
+  ARGV.delete('--single')
+  single = true
+end
+
+if ARGV.include?('--sequential')
+  ARGV.delete('--sequential')
+  sequential = true
+end
+
 jurisdiction = ARGV.shift
 if jurisdiction.nil? || !jurisdiction.match(/^[A-Z][A-Z]$/)
   puts "You must provide the jurisdiction code (e.g., MA) as the first argument"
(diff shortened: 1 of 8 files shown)


tools/manual_api_testing/push.rb
```diff
@@ -46,16 +56,25 @@
 
 # Submit the files in chunks of 20
 failures = []
-Parallel.each(files.each_slice(20)) do |slice|
+options = {}
+if sequential # Don't use parallel processing
+  options[:in_processes] = 0
+end
+slice_size = single ? 1 : 20
+Parallel.each(files.each_slice(slice_size), options) do |slice|
```
holmesie, line 64: Something I've been trying to do more of lately when writing Ruby scripts that depend on gems is to include a section before the requires; such as:

```
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'oauth2'
  gem 'activesupport'
  gem 'parallel'
end
```

Makes it a bit easier to run these sorts of scripts if the developer doesn't have these gems already installed.

pkmitre: Nice, thanks!

smacadam (approved):

PR #335, opened 2022-11-21
Improve handling of middle names in IJE

pkmitre:
This pull request

1. Adds a major test that goes through each field and makes sure that no other field stomps on its value when being set
2. Updates how middle names are handled in response to the results of that test:
    1. raising an exception if a middle name is set before a first name in IJEMortality
    2. returns the correct length strings for middle names in IJEMortality
3. Adding further tests to cover the expected behavior of middle names
4. Incidentally fixes the display order of IJE fields in the CLI

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,19 @@ deathRecord.DeathDay = -1;
 deathRecord.DeathTime = "-1";
 ```
 
+#### Names in FHIR
+
+FHIR manages names in a way that there is a fundamental incompatibility with IJE: in FHIR the
+"middle name" is stored as the second element in an array of given names. That means that it's not
+possible to set a middle name without first setting a first name. The library handles this by
+
+1. Requiring the entire given name (first name and any middle names) to be set all at once when
+using the DeathRecord class
+2. Raising an exception if a middle name is set before a first name when using the IJEMortality
+class
+3. Resetting the middle name if the first name is set again when using the IJEMortality class;
+setting the first name and then the middle name ensures no issues will occur.
+
 #### Helper Methods for Value Sets
 
 For fields that contain coded values it can involve some extra effort to provide the code, the code
--- a/VRDR.CLI/Program.cs
+++ b/VRDR.CLI/Program.cs
@@ -477,7 +477,7 @@ namespace VRDR.CLI
                 DeathRecord d = new DeathRecord(File.ReadAllText(args[1]));
                 IJEMortality ije1 = new IJEMortality(d, false);
                 // Loop over every property (these are the fields); Order by priority
-                List<PropertyInfo> properties = typeof(IJEMortality).GetProperties().ToList().OrderBy(p => p.GetCustomAttribute<IJEField>().Priority).ToList();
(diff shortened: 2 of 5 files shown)


saulakravitz (approved):

VRDR.CLI/Program.cs
```diff
@@ -762,15 +762,15 @@ record = coding.DeathRecord;
                         var d = submission.DeathRecord;
                         IJEMortality ije1 = new IJEMortality(d, false);
                         // Loop over every property (these are the fields); Order by priority
-                        List<PropertyInfo> properties = typeof(IJEMortality).GetProperties().ToList().OrderBy(p => p.GetCustomAttribute<IJEField>().Priority).ToList();
+                        List<PropertyInfo> properties = typeof(IJEMortality).GetProperties().ToList().OrderBy(p => p.GetCustomAttribute<IJEField>().Location).ToList();
```
smacadam, line 765: Would this mean the priority values that we use for the death record fields get ignored? Does that matter?

VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -3384,6 +3384,190 @@ public void Get_EmptyLiteralRaceFields()
...
+                { "CERTZIP", "17505" },
+                { "CERTDATE", "11142020" },
+                { "DTHCOUNTRYCD", "US" },
+                { "DTHCOUNTRY", "United States" },
+                { "PLACE1_1", "H" },
+                { "PLACE1_2", "I" },
+                { "PLACE8_1", "Hi 8_1" },
+                { "PLACE20", "Hi 20_1"}
+            };
+            // For each field, create a record, set that field, set all the other fields, and make sure the first field still has the same value
```
smacadam, line 3555: I remember you said that setting the first name will clear a middle name, is middle name left out of this test?

README.md
```diff
@@ -172,6 +172,19 @@ deathRecord.DeathDay = -1;
 deathRecord.DeathTime = "-1";
 ```
 
+#### Names in FHIR
+
+FHIR manages names in a way that there is a fundamental incompatibility with IJE: in FHIR the
+"middle name" is stored as the second element in an array of given names. That means that it's not
+possible to set a middle name without first setting a first name. The library handles this by
+
+1. Requiring the entire given name (first name and any middle names) to be set all at once when
```
smacadam, line 181: I expected there to be some changes in one of the DeathRecord files to enforce this, am I missing it?

VRDR.CLI/Program.cs
saulakravitz: The priority is used in Canary (I think), but for dumping the IJE values we really want the dump to be in the same order as the fields in the IJE record, and that is Location.

VRDR.Tests/DeathRecord_Should.cs
saulakravitz: The comment says:
// MNAME, DMIDDLE, DDADMID, DMOMMID, SPOUSEMIDNAME, CERTMIDDLE (middle names behave oddly due to how FHIR represents names)

So, yes, these are left out, since we already know they are poorly behaved.   Pete added tests that are expected to fail for these.

PR #344, opened 2022-12-06
Helper Properties for AgeAtDeath

johnarleyburns:
1) Changed 'unit' to 'code' for AgeAtDeath
2) Added AgeAtDeath* helper methods for years, months, days, hours, minutes
3) Added test methods
4) Updates as per PR feedback

Committer: John Arley Burns <johnburns@ctex-inc.com>


VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -5416,31 +5419,197 @@ public string DeathLocationTypeHelper
...
+        {
+            get {
+                if (AgeAtDeathObs?.Value != null)
+                {
+                    Quantity quantity = (Quantity)AgeAtDeathObs.Value;
+                    if (quantity.Code == "a")
+                    {
+                        return Convert.ToInt32(quantity.Value);
+                    }
+                }
```
pkmitre, line 5450: This could probably be simplified by using AgeAtDeath, e.g.,

```
if (AgeAtDeath["code"] == "a")
{
    return AgeAtDeath["value"];
}
```

VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -5416,31 +5419,197 @@ public string DeathLocationTypeHelper
...
+                return 0; // not sure how to best handle this case
```
pkmitre, line 5451: Does it it make sense to have

```
public int? AgeAtDeathYears
{
    ...
    return null;
}
```

In that case the setter should check for null

VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -5416,31 +5419,197 @@ public string DeathLocationTypeHelper
...
+            {
+                AgeAtDeath = new Dictionary<string, string>() {
+                    { "value", Convert.ToString(value) },
+                    { "code", "min" }
+                };
+            }
+        }
+
         /// <summary>Decedent's Age At Death Edit Flag.</summary>
-        /// <value>the decedent's age at death edit flag. A Dictionary representing a code, containing the following key/value pairs:
```
pkmitre, line 5424: These updates to the comments for AgeAtDeathEditFlag should not be included

VRDR/IJEMortality.cs
```diff
@@ -1153,8 +1153,6 @@ public string AGETYPE
                         // using the supplied options and return
                         Dictionary<string, string> dict = new Dictionary<string, string>();
                         dict.Add("code", fhirValue);
-                        dict.Add("unit", ValueSets.UnitsOfAge.Codes[i, 1]);
-                        dict.Add("system", ValueSets.UnitsOfAge.Codes[i, 2]);
```
pkmitre, line 1157: It would be helpful to leave the above two lines in.

pkmitre (changes requested):

pkmitre (approved):

PR #346, opened 2022-12-12
Add lookup table for endpoints

smacadam:
Adding a lookup table to the library so we can fill in missing source endpoints in the API. The values are set according to source endpoints jurisdictions have submitted so far to NCHS and can be updated in the future.

This change is required before the 471 branch can be merged into the Reference API


smacadam:
Checked with Veronique and she does not do string validation on the source endpoint, just checks that the field is populated. The values in this table can be a best guess and verified by jurisdictions. At this time, neither the NCHS API nor STEVE use the source endpoint value for any routing purposes, so the consequence of one of the endpoints being incorrect is currently very low.

PR #347, opened 2022-12-16
Empty strings that should be null/not present for a blank death record

RobertScalfani:
This PR fixes cases where the properties of a blank death record should have `null` or no value rather than blank whitespace.

### DIFF
--- a/VRDR.Tests/DeathRecord_Should.cs
+++ b/VRDR.Tests/DeathRecord_Should.cs
@@ -3432,7 +3432,7 @@ namespace VRDR.Tests
             Assert.Equal("000000000042", mortalityrosterbundle.StateLocalIdentifier1);
             Assert.Equal("100000000001", mortalityrosterbundle.StateLocalIdentifier2);
             Assert.Equal("", mortalityrosterbundle.CertificationRole["code"]); // should be empty
-            Assert.Equal(null, mortalityrosterbundle.PregnancyStatusHelper); // should be missing
+            Assert.Null(mortalityrosterbundle.PregnancyStatusHelper); // should be missing
             // TODO: Fill out tests
         }
         [Fact]
--- a/VRDR.Tests/DeathRecord_Should.cs
+++ b/VRDR.Tests/DeathRecord_Should.cs
@@ -3432,7 +3432,7 @@ namespace VRDR.Tests
             Assert.Equal("000000000042", mortalityrosterbundle.StateLocalIdentifier1);
             Assert.Equal("100000000001", mortalityrosterbundle.StateLocalIdentifier2);
             Assert.Equal("", mortalityrosterbundle.CertificationRole["code"]); // should be empty
-            Assert.Equal("", mortalityrosterbundle.PregnancyStatusHelper); // should be missing
+            Assert.Equal(null, mortalityrosterbundle.PregnancyStatusHelper); // should be missing
             // TODO: Fill out tests
         }
         [Fact]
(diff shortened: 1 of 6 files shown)


VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -3315,7 +3316,7 @@ public void Test_MortalityRosterBundle()
             Assert.Equal("000000000042", mortalityrosterbundle.StateLocalIdentifier1);
             Assert.Equal("100000000001", mortalityrosterbundle.StateLocalIdentifier2);
             Assert.Equal("", mortalityrosterbundle.CertificationRole["code"]); // should be empty
-            Assert.Equal("", mortalityrosterbundle.PregnancyStatusHelper); // should be missing
+            Assert.Equal(null, mortalityrosterbundle.PregnancyStatusHelper); // should be missing
```
saulakravitz, line 3319: per Github auto-comment:  should this be Assert.Null(mortality...) instead?

VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -2091,6 +2136,11 @@ public string DateOfBirth
                     Extension withinCityLimits = new Extension();
                     withinCityLimits.Url = ExtensionURL.WithinCityLimitsIndicator;
                     withinCityLimits.Value = DictToCoding(value);
+                    // Coding coding = DictToCoding(value);
```
smacadam, line 2139: Does the DictToCoding function handle setting the null value correctly?

remove commented out code if not needed

VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -2489,6 +2489,7 @@ public string EthnicityLiteral
             }
             set
             {
+                // TODO: This setter should not do anything if the value is null or an empty string
```
smacadam: should this TODO be addressed now or does it need some clarification?

VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -2496,7 +2556,10 @@ public string EthnicityLiteral
                 InputRaceAndEthnicityObs.Component.RemoveAll(c => c.Code.Coding[0].Code == NvssEthnicity.Literal);
                 Observation.ComponentComponent component = new Observation.ComponentComponent();
                 component.Code = new CodeableConcept(CodeSystems.ComponentCode, NvssEthnicity.Literal, NvssEthnicity.Literal, null);
-                component.Value = new FhirString(value);
+                if (!String.IsNullOrEmpty(value))
```
smacadam, line 2559: For fields like this that create an observation as part of the setter, I think the IsNullOrEmpty check should be wrapped around all of the code in the setter. Check with Saul to see what the best practice is for null values in fhir

VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -353,11 +362,15 @@ public string StateSpecific
...
+                return null;
             }
             set
             {
                 // TODO: Handle case where Composition == null (either create it or throw exception)
+                if (String.IsNullOrWhiteSpace(value))
+                {
+                    return;
+                }
                 Composition.Extension.RemoveAll(ext => ext.Url == ExtensionURL.StateSpecificField);
```
smacadam, line 374: From Saul's email, fields like State Specific that have a RemoveAll or some sort of delete that clears the field should do the delete first and then check whether the value is null and create the extension if its not null or white space.

VRDR/DeathRecord_submissionProperties.cs
```diff
@@ -5652,15 +5652,15 @@ private void _setAgeAtDeathForCode(string code, int? value)
...
-                if (IsDictEmptyOrDefault(value) && AgeAtDeathObs == null)
-                {
-                    return;
-                }
                 if (AgeAtDeathObs == null)
                 {
                     CreateAgeAtDeathObs();
                 }
                 AgeAtDeathObs.Value.Extension.RemoveAll(ext => ext.Url == ExtensionURL.BypassEditFlag);
+                if (IsDictEmptyOrDefault(value) && AgeAtDeathObs == null)
```
smacadam, line 5660: since you always create the AgeAtDeathObs, this will never be true

smacadam (approved):

PR #379, opened 2023-02-10
Issue 359 fixes for Connectathon 

johnarleyburns:
* Connectathon test record fixes
* DateOfDeath valueTime formatting correction
* Certifier allowed to be empty
* LOINC code display value corrected to "Location of Death"
* NVSS Race and Ethnicity display values corrected
* Test methods added to verify each of the above


RichDCTEX (approved):

VRDR/RaceAndEthnicity.cs
```diff
@@ -72,6 +74,29 @@ public static List<string> GetBooleanRaceCodes()
...
+                case GuamanianOrChamorro:
+                    return "Guamanian Or Chamorro";
+                case OtherPacificIslander:
+                    return "Other Pacific Islander";
+                case OtherRace:
+                    return "Other Race";
+                default:
+                    return code;
+            }
+        }
```
saulakravitz, line 99: Is this used anywhere?

saulakravitz (commented):
See my minor comments.

johnarleyburns:
VRDR/DeathRecord_util.cs: I can check if the value is 5 characters and add
":00" to the end if so, would you prefer that?

VRDR/RaceAndEthnicity: yes this section is used in a unit test and appears
in json output.

VRDR/RaceAndEthnicity.cs
johnarleyburns: Yes in testing and appears in json output.

VRDR/DeathRecord_util.cs
```diff
@@ -120,6 +121,12 @@ private void SetPartialTime(Extension partialDateTime, String value)
             part.Extension.RemoveAll(ext => ext.Url == OtherExtensionURL.DataAbsentReason);
             if (value != null && value != "-1")
             {
+                // we need to force it to be 00:00:00 format to be compliant with the IG because the FHIR class doesn't
+                if (value.Length < 8)
+                {
```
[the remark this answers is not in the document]
johnarleyburns: Yes I think we can refactor to more directly handle time setting errors in the future to better make sure not only is it the right length but has properly formatted values, and identifies an error if the wrong format is used.

PR #386, opened 2023-02-16
374 fix and test decedent education level in the composition

RichDCTEX:
Fixed code, created test that proves out the fix.  If you undo the fix the test fails, so the testing seems correct.

### DIFF
--- a/VRDR.Tests/DeathRecord_Should.cs
+++ b/VRDR.Tests/DeathRecord_Should.cs
@@ -3836,12 +3836,12 @@ namespace VRDR.Tests
         public void TestEducationLevelObs()
         {
             var testRecord = new DeathRecord();
-            Composition composition = testRecord.ReturnCompositionForTesting();
+            Composition composition = testRecord.GetComposition();
             bool covered = false;
 
             int[] beforeCounts = new int[composition.Section.Count];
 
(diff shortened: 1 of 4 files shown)


johnarleyburns (changes requested):
Please add the [Fact] attribute to method TestAutopsyAvailableCodes

VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -3833,6 +3833,33 @@ public void TestPregnancyCodes()
...
+            {
+                beforeCounts[i] = composition.Section[i].Entry.Count;
+            }
+
+            testRecord.EducationLevelHelper = "HS";
+
+            // Verify we added a section entry with the above statement
+            for (int i = 0; i < beforeCounts.Length; i++)
+            {
+                if(beforeCounts[i]+1 == composition.Section[i].Entry.Count)
```
johnarleyburns, line 3854: Are you saying at least one section[i] has an extra entry?  If so you should "break" on this condition instead of continuing looping.

Also, you should check the section entry and make sure it's for DecedentDemographics by name, and not some other name.

johnarleyburns (changes requested):

johnarleyburns (approved):

PR #396, opened 2023-03-27
Add filtering logic

brynneprince:


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Changelog
 
+### v4.0.0.preview19-with-filtering - 2023-03-16
```
johnarleyburns, line 3: Question: will you also make these same changes to the VRDR Java library?  Some jurisdictions are using the Java library.

VRDR.Tests/Filtering_Should.cs
```diff
@@ -0,0 +1,341 @@
...
+                {
+                    passedFields.Add(property.Name);
+                    
+                    // Console.WriteLine($"{property.Name}");
+                    // Console.WriteLine(preFilteredObject);
+                    // Console.WriteLine(postFilteredObject);
+                }
+            }
+            
+            Console.WriteLine("=============Field filtering test=============");
```
johnarleyburns, line 129: There are no assertions here, this test needs to have assertions so that it fails if the filtered passed and failed fields are not what is expected.  The same for the subsequent tests which are lacking assertions.

VRDR.Tests/Filtering_Should.cs
```diff
@@ -0,0 +1,341 @@
...
+                    throw new Exception(e.Message);
+                }
+            }
+            
+            Console.WriteLine("=============Jurisdiction filtering test=============");
+            Console.WriteLine($"Passed jurisdiction count = {passedJurisdictions.Count} ({JsonConvert.SerializeObject(passedJurisdictions)})");
+            Console.WriteLine($"Failed jurisdiction count = {failedJurisdictions.Count} ({JsonConvert.SerializeObject(failedJurisdictions)})");
+        }
+        
+        private static DeathRecord getDeathRecordFromMessage(BaseMessage message)
```
johnarleyburns, line 314: Parsing already extracts the record, as in VRDR.CLI/Program.cs "extract" option code below.  Suggest adding DeathRecordUpdateMessage to this code below, refactoring it to the BaseMessage class, then remove "getDeathRecordFromMessage()" here and use the refactored BaseMessage method.

            else if (args.Length == 2 && args[0] == "extract")
            {
                BaseMessage message = BaseMessage.Parse(File.ReadAllText(args[1]));
                DeathRecord record;
                switch (message)
                {
                    case DeathRecordSubmissionMessage submission:
                        record = submission.DeathRecord;
                        Console.WriteLine(record.ToJSON());
                        break;
                    case CauseOfDeathCodingMessage coding:
                        record = coding.DeathRecord;
                        Console.WriteLine(record.ToJSON());
                        break;
                    case DemographicsCodingMessage coding:
                        record = coding.DeathRecord;
                        Console.WriteLine(record.ToJSON());
                        break;
                }
                return 0;
            }

VRDR/DeathRecord.csproj
```diff
@@ -11,6 +11,7 @@
     <Authors>aholmes@mitre.org</Authors>
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <LangVersion>latest</LangVersion>
```
johnarleyburns, line 14: This does not appear to be necessary as it comes from the framework, please remove.

doc/Messaging.md
johnarleyburns, line 1: Please add documentation about filtering similar to how other features are documented in this file.

johnarleyburns (changes requested):
Looks good, a lot of excellent code.  Only a few issues to resolve:
1. Refactor extraction of death record to BaseMessage class
2. Add assertions to all filtering test cases where they are missing
3. Include documentation on using the new filtering service
4. Question: Is this service going to be included in the Java library as well?

CHANGELOG.md
brynneprince: I can do, I think the idea, for now, is to get the filtering logic onto the .net VRDR library and get some assistance with our filtering logic.
We had issues previously and decided to remove the filtering from our code to the .net VRDR library to allow for other parties to view this logic easily. Once we have got confirmation that the logic is good, then filtering can be used from this library.
At that point I'll make these changes to the java VRDR library.

VRDR.Tests/Filtering_Should.cs
brynneprince: For some of these tests, we aren't testing a single assertion. Some test that a valid BaseMessage is generated from the filter.cs class. I see your point with the ones that have success and failure arrays, I will add them there.
What would you suggest for the test that test multiple assertions?

VRDR.Tests/Filtering_Should.cs
brynneprince: I just want to confirm what you mean by this comment.
Add the case statement for DeathRecordUpdateMessage.
Remove the method, getDeathRecordFromMessage() completely.
I am uncertain about refactoring to the BaseMessage class and how to use this logic from the test class.

CHANGELOG.md
johnarleyburns: Okay we will consider this a C# library change only for now, I've created an issue to track the Java change in the future: https://github.com/nightingaleproject/VRDR_javalib/issues/14

VRDR.Tests/Filtering_Should.cs
johnarleyburns: You can add all the assertions in the single test, so that if any assertions fail, the whole test fails.  For instance in Messaging_Should.cs you can see most of the tests have many assertions per test.  For instance you could do individual field assertions for important fields, and total field count checks for everything.

For the BaseMessage.Parse runs, you can do:
1. Remove the try/catch blocks as they shouldn't be necessary around the test, for instance see Messaging_Should::CreateUpdateFromJson()
2. Add an assertion about the returned message, such as that it isn't null, has the correct certificate number, just something to tell you the message was parsed properly after the filter service processed it

For at least one test, make sure the filtering fails on an invalid value, such as is done with Messaging_Schold.SelectMessageType() where it does Assert.Throws() calls to make sure invalid messages generate exceptions on parsing, for filtering you want something similar.

Because we have a number of upstream consumers of the library, we do more than the average amount of tests to give higher confidence in the released software.

johnarleyburns:
Currently the certificate number is only validated in the Message
Parameters section as described in the VRDR Messaging IG:

http://build.fhir.org/ig/nightingaleproject/vital_records_fhir_messaging_ig/branches/main/StructureDefinition-VRM-MessageParameters.html

The certificate number should be an unsigned integer of up to six
characters, so 0..999999 inclusive (although 1 may be a better starting
number as 0 might throw an error somewhere in the pipe).  For instance this
example Parameters section from the IG with the certificate number:

{
  "resourceType" : "Parameters",
  "id" : "Parameters-Example1",
  "meta" : {
    "profile" : [
      "
http://cdc.gov/nchs/nvss/fhir/vital-records-messaging/StructureDefinition/VRM-MessageParameters
"
    ]
  },
  "parameter" : [
    {
      "name" : "jurisdiction_id",
      "valueString" : "NY"
    },
    {
      "name" : "cert_no",
      "valueUnsignedInt" : 123456
    },
    {
      "name" : "death_year",
      "valueUnsignedInt" : 2018
    },
    {
      "name" : "state_auxiliary_id",
      "valueString" : "abcdef10"
    }
  ]
}

The applicable part of the C# library is in BaseMessage::CertNo:

        public uint? CertNo
        {
            get
            {
                return
(uint?)Record?.GetSingleValue<UnsignedInt>("cert_no")?.Value;
            }
            set
            {
                Record.Remove("cert_no");
                if (value != null)
                {
                    if (value > 999999)
                    {
                        throw new ArgumentException("Certificate number
must be a maximum of six digits");
                    }
                    Record.Add("cert_no", new UnsignedInt((int)value));
                }
            }
        }

This is the only place the validation is being done by the library itself
on parsing the message, the NVSS API code does its own additional handling.

brynneprince:
Thank you for that.

VRDR.Tests/Filtering_Should.cs
brynneprince: Did you see my comment here?

johnarleyburns (changes requested):

VRDR.CLI/Program.cs
```diff
@@ -1101,6 +1104,32 @@ record = coding.DeathRecord;
...
+                string filteredFile;
+                
+                try
+                {
+                    filteredFile = FilterService.filterMessage(baseMessage).ToJson();
+                    BaseMessage.Parse(filteredFile);
+                    Console.WriteLine($"File successfully filtered");
+                    
+                    string filteredFilePath = "./VRDR.Filter/filteredFile.json";
+                    File.WriteAllText(filteredFilePath, filteredFile);
```
johnarleyburns, line 1124: Please make the filtered file output a configurable file in args[2] so that it doesn't overwrite sections of the source code tree.

VRDR.CLI/Program.cs
```diff
@@ -1101,6 +1104,32 @@ record = coding.DeathRecord;
...
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine($"Error occured while filtering file ({e})");
```
johnarleyburns, line 1128: If an error occurs, please just let the exception be thrown so the full stack trace shows, remove the try/catch block.

johnarleyburns (changes requested):

brynneprince:
Another question regarding the certificate number validation. How do I trigger the set method for the certificate number? I would have assumed calling a parse method to serialise the message into a base message would trigger that. If you could point me in the right direction, that would be greatly appreciated.

brynneprince:
We are doing exactly what the NVSS API is doing.
So before we handle a bundle or single record, we perform this, `Bundle bundle = BaseMessage.ParseGenericBundle(file, true);`.
And within each case, we the first thing we do before handling a message is this, `BaseMessage message = BaseMessage.Parse(text.ToString());`.

johnarleyburns:
Could you help me understand more, is there a case where the CertNo is not
getting set, and if so I can help diagnose (you can reach me at
***@***.***).

johnarleyburns (approved):
Very nice, great code addition, well written and with extensive tests and documentation.

brynneprince:
Your email was redacted from the comment.

Issue #385, opened 2023-02-16
Add Support for PartialDateTime for PPTIME/PPDATESIGNED /DeathDate


saulakravitz:
STU2.1 updates the DeathDate.component[datetimePronouncedDead].valueDateTime to support a PartialDateTime extension.
The code for DeathDate needs to be updated to support this
This eliminates some wonky code that created a placeholder date if PPTIME was assigned and PPTIME wasn't, since the dateTime field must contain a date.

http://build.fhir.org/ig/HL7/vrdr/branches/master/StructureDefinition-vrdr-death-date.html

PR #398, opened 2023-03-31
Add Support for PartialDateTime for PPTIME/PPDATESIGNED /DeathDate #3…

johnarleyburns:
Add Support for PartialDateTime for PPTIME/PPDATESIGNED /DeathDate #385

1. Uncertain about how time zone datetime source should work
2. Uncertain about how to handle partial date times in IJE with "-1" / "99" values vs. just not generating the value at all


pkmitre:
First, a question: why have date and time pronounced been converted to use partial date time? The reason they were used for other fields is that IJE allowed, say, a year and month without a day for those fields and we needed to be able to support that for backwards compatibility. That's not the case for date and time pronounced, and so this change actually makes it more difficult to align the IJE and FHIR. See below...

Regarding your comment "Uncertain about how time zone datetime source should work": IJE does not include time zone information, and the assumption is that the record specifies times using the time zone of the submitting jurisdiction.

Regarding your comment "Uncertain about how to handle partial date times in IJE with "-1" / "99" values vs. just not generating the value at all": This is where the IJE/FHIR conversion issue mentioned above comes into play. Since there are no individual year, month, and day fields for date pronounced, there is no concept of "99". Since there is no concept of "99", the concept of "-1" doesn't make sense either. How do you map a FHIR record that has pronounced year set to "2023", pronounced month set to "-1", and pronounced day set to "1" to IJE?

johnarleyburns:
I like that approach too, let me take a crack at it and I'll come up with a
revised PR.

saulakravitz:
I will update the master branch of the IG.
________________________________
From: John Burns ***@***.***>
Sent: Monday, April 24, 2023 12:37 PM
To: nightingaleproject/vrdr-dotnet ***@***.***>
Cc: Saul A Kravitz ***@***.***>; Review requested ***@***.***>
Subject: [EXT] Re: [nightingaleproject/vrdr-dotnet] Add Support for PartialDateTime for PPTIME/PPDATESIGNED /DeathDate #3… (PR #398)


I like that approach too, let me take a crack at it and I'll come up with a
revised PR.

VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -3076,21 +3076,141 @@ public void Get_DateOfDeath_Timezone()
...
+            d.DateOfDeathPronouncementTime = "-1";
+            Assert.Equal(2022, d.DateOfDeathPronouncementYear);
+            Assert.Equal(-1, d.DateOfDeathPronouncementMonth);
+            Assert.Null(d.DateOfDeathPronouncementDay);
+            Assert.Equal("-1", d.DateOfDeathPronouncementTime);
+            IJEMortality ije = new IJEMortality(d, false);
+            Assert.Equal("99992022", ije.PPDATESIGNED);
+            Assert.Equal("9999", ije.PPTIME);
+        }
+*/
```
saulakravitz, line 3164: Suggest deleting

VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -3076,21 +3076,141 @@ public void Get_DateOfDeath_Timezone()
...
-            Assert.Equal("2018-02-20T16:48:06-05:00", DeathRecord1_JSON.DateOfDeathPronouncement);
-            Assert.Equal("2020-11-13T16:39:40-05:00", DeathCertificateDocument2_JSON.DateOfDeathPronouncement);
-            Assert.Equal("2019-02-20T16:48:06-05:00", DeathRecord1_XML.DateOfDeathPronouncement);
+            Assert.Equal("2018-02-20T16:48:06", DeathRecord1_JSON.DateOfDeathPronouncement);
+            Assert.Equal("2020-11-13T16:39:40", DeathCertificateDocument2_JSON.DateOfDeathPronouncement);
+            Assert.Equal("2019-02-20T16:48:06", DeathRecord1_XML.DateOfDeathPronouncement);
         }
 
+        [Fact]
+        public void Get_DateOfDeathPronouncement_PPDATESIGNED_and_PPTIME_Roundtrip()
```
saulakravitz, line 3096: I like the new test cases.

VRDR.Tests/DeathRecord_Should.cs
johnarleyburns: Done

VRDR.Tests/DeathRecord_Should.cs
johnarleyburns: Currently failing on the timezone difference, as the code is assuming it's running in Zulu time, not sure how this work when importing dates and times from IJE vs from JSON vs the timezone returned by defaut for DateOfDeathPronouncement.

johnarleyburns: IJE conversion now working as well, new tests added.

consultantusa (approved):

PR #413, opened 2023-05-09
add leading zeroes to IJEFields 'AUXNO2' and 'AUXNO'

consultantusa:


VRDR/IJEMortality.cs
```diff
@@ -2971,6 +2972,7 @@ public string AUXNO2
             {
                 if (!String.IsNullOrWhiteSpace(value))
                 {
+					value = value.PadLeft(12 , '0');
```
johnarleyburns, line 2975: Please fix the spacing, set your editor to convert tabs to spaces so it's consistent.

VRDR/IJEMortality.cs
```diff
@@ -937,7 +937,8 @@ public string AUXNO
             {
                 if (!String.IsNullOrWhiteSpace(value))
                 {
-                    LeftJustified_Set("AUXNO", "StateLocalIdentifier1", value);
+                    value = value.PadLeft(12 , '0');
+					LeftJustified_Set("AUXNO", "StateLocalIdentifier1", value);
```
johnarleyburns, line 941: Please fix the spacing, set your editor to convert tabs to spaces so it's consistent.

PR #416, opened 2023-05-10
403 add leading zero handling for auxno1 and auxno2 analagous to filo

consultantusa:


VRDR/IJEMortality.cs
johnarleyburns, line 1: You'll need to update the test cases as well so they pass with the new zero-padded results.

johnarleyburns (changes requested):

VRDR/IJEMortality.cs
consultantusa: I was looking at DeathRecordTests.csproj based on error message and just saw a number of json files
image

johnarleyburns: Scroll up to see the full list of test errors

PR #421, opened 2023-05-15
update 3 json records

consultantusa:


VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -3559,12 +3559,12 @@ public void CheckConnectathonRecordRaceDisplayValue()
...
             DeathRecord dr1 = VRDR.Connectathon.FideliaAlsup();
-            Assert.Equal("62", dr1.AgeAtDeath["value"]);
+            Assert.Equal("63", dr1.AgeAtDeath["value"]); //62
             Assert.NotNull(dr1.ToDescription()); // This endpoint is used by Canary
             IJEMortality ije = new IJEMortality(dr1, false); // Don't validate since we don't care about most fields
-            Assert.Equal("062", ije.AGE);
+            Assert.Equal("063", ije.AGE); //62
             Assert.Equal("478151044", ije.SSN);
-            Assert.Equal("Unrestrained ejected driver in rollover motor vehicle accident", ije.HOWINJ.Trim());
+            Assert.Equal(, ije.HOWINJ.Trim()); //Unrestrained ejected driver in rollover motor vehicle accident
```
johnarleyburns, line 3567: You should just remove this missing test case

VRDR.Tests/DeathRecord_Should.cs
```diff
@@ -3747,7 +3747,7 @@ public void TestForOverwrites()
                 // This list of fields is fairly comprehensive, though some have been intentionally left out:
                 // STATETEXT_D, STATEBTH, and FUNSTATE (setting these are a no-ops)
                 // MNAME, DMIDDLE, DDADMID, DMOMMID, SPOUSEMIDNAME, CERTMIDDLE (middle names behave oddly due to how FHIR represents names)
-                { "DOD_YR", "2022" },
+                { "DOD_YR", "2023"}, //2022" },
```
johnarleyburns, line 3750: Remove the old value comments like "// 2022" and "//2"

VRDR/Connectathon.cs
```diff
@@ -106,12 +106,12 @@ public static DeathRecord TwilaHilty()
...
             ije.AUTOPF = "X";
             ije.TOBAC = "U";
-            ije.PREG = "2";
+            ije.PREG = "8";//2";
             ije.PREG_BYPASS = "0";
             ije.CERTL = "D";
-            ije.SUR_MO = "01";
-            ije.SUR_DY = "10";
-            ije.SUR_YR = "2022";
+            ije.SUR_MO = null;
```
johnarleyburns, line 112: You don't need to set to null, just remove these lines so it remains the default value

VRDR/Connectathon.cs
```diff
@@ -194,17 +194,17 @@ public static DeathRecord FideliaAlsup()
...
             ije.TOBAC = "U";
-            ije.PREG = "8";
+            ije.PREG = "1";//8";
             ije.PREG_BYPASS = "0";
-            ije.DOI_MO = "03";
-            ije.DOI_DY = "16";
-            ije.DOI_YR = "2022";
-            ije.TOI_HR = "1015";
-            ije.WORKINJ = "Y";
+            ije.DOI_MO=null;// = "03";
```
johnarleyburns, line 203: Instead of setting to null, remove the lines so the default value is used

johnarleyburns (changes requested):

johnarleyburns:
Please also check the three Connectathon records in the JSON test files to
make sure we don’t have any invalid values there (search by First and Last
Name, e.g. Twila)

PR #429, opened 2023-05-18
Update IJEMortality.cs

consultantusa:


johnarleyburns (changes requested):
Please add test case for these exceptions.

johnarleyburns (changes requested):
Please add test case for these exceptions.

consultantusa:
Added test case. Thank you.

johnarleyburns (approved):

PR #447, opened 2023-06-07
443-add-transportrolehelper-but-with-correct-fhir-mapping

consultantusa:


johnarleyburns (changes requested):
Good only small changes accidentally reverted need to be removed from diff.

johnarleyburns:
Looks good only change needed is a few places where lines were reverted not
related to transportation role, such as certifier, mortician.

johnarleyburns (approved):

PR #456, opened 2023-06-12
fix: yes-no value set url

whichwit:
Should be `http://terminology.hl7.org/CodeSystem/v2-0136`.


johnarleyburns (approved):

johnarleyburns:
Looks good, I’ve approved, but held off merging in case there were any
other files you wanted to change also.

PR #459, opened 2023-06-14
feat: allow retrieval of all Connectathon records

whichwit:
Add new static method All() to return all available pre-set records.


johnarleyburns (changes requested):
Two changes:
1. Create an issue and map this branch to the issue
2. Create a unit test for the new function

johnarleyburns (approved):

PR #464, opened 2023-06-19
452-add-messagebundle-as-a-read-only-public-property-to-deathrecord-c…

consultantusa:


johnarleyburns (changes requested):
The git history in this PR is incorrect, I suggest you abandon this PR and create a new PR to start from scratch.

PR #469, opened 2023-06-22
452 add messagebundle as a read only public property to deathrecord class 1

consultantusa:


johnarleyburns (changes requested):
Remove changes made to DeathRecord_constructors.cs, instead just change BaseMessage.cs and add a test to access MessageBundle.

VRDR.Tests/Messaging_Should.cs
```diff
@@ -148,6 +148,13 @@ public void CreateSubmissionFromBundle()
             Assert.Equal(submission.NCHSIdentifier, parsed.NCHSIdentifier);
         }
 
+		[Fact]
+        public void ProtectedSetMessageBundle()
+        {
+            DeathRecordSubmissionMessage submission = new DeathRecordSubmissionMessage();
+            Assert.NotNull(submission);
+		}
```
johnarleyburns, line 156: Add a submission.MessageBundle not null assertion here, and reformat the right bracket whitespace.

johnarleyburns (changes requested):

johnarleyburns (changes requested):
Looks good only needs readme update now

johnarleyburns (approved):

PR #477, opened 2023-06-29
docs: add DeathRecord.cs annotations and library developers section to README

consultantusa:


whichwit:
It's `docs:` instead of `doc:`.

whichwit (dismissed):
- Sample code blocks should be wrapped between triple quotes (''') in the Markdown
- Inline code should be wrapped in between backquotes (`)
- Change type should be `docs` in the PR title; remove PR body if the content isn't relevant

johnarleyburns (changes requested):
In addition to Mr. Yang's suggestions, also please add documentation on the IJEField annotation in IJEMortality.cs, because this also affects how the DeathRecord(ije) constructor works.

PR #482, opened 2023-07-07
feat: add setter/getter for Decedents gender

consultantusa:


saulakravitz (commented):
I defer to @pkmitre , but it looks good to me.
Please wait for @pkmitre to review.

Issue #362, opened 2023-01-26
Generate Errors For Poorly Formulated Submissions -- Partial Date and Partial Date Time


saulakravitz:
We noted in trying to validate submissions from jurisdictions there were several pathologies relating to the PartialDate and PartialDateTime extensions in the IG.   Specifically:
1) Incorrected structured PartialDate/PartialDateTime Extensions:
   - incorrect URLs for the sub-extensions (DateDay, DateTime, DateMonth, DateYear).
   - failure to include a value for the extensions
    - including an extension with the sub-extensions, instead of a value
 2) Inclusion of a PartialDate when a PartialDateTime is appropriate (and vice versa)
 
 Code should be added to check for these pathologies, and report them as an error.

PR #492, opened 2023-07-14
fix: partialDate and partialDateTime validation

RobertScalfani:
This PR adds validation checks for PartialDate and PartialDateTime extensions for an input FHIR Death Record. If any of the required [Date-Day, Date-Month, Date-Year, Date-Time] sub-extensions are missing for a `http://hl7.org/fhir/us/vrdr/StructureDefinition/PartialDate` or `PartialDateTime`, it will throw an error.
Also, if there are any extra invalid sub-extensions, it will throw an error.

This covers GitHub issue https://github.com/nightingaleproject/vrdr-dotnet/issues/362 and is a result from our group discussion a few weeks ago.

### DIFF
--- a/VRDR/DeathRecord_util.cs
+++ b/VRDR/DeathRecord_util.cs
@@ -1075,6 +1075,51 @@ namespace VRDR
                 names.Add(name);
             }
         }
+
+        private static void ValidatePartialDates(Bundle bundle)
+        {
+            System.Text.StringBuilder errors = new System.Text.StringBuilder();
+            List<DomainResource> resources = bundle.Entry.Select(entry => (DomainResource) entry.Resource).ToList();
+
+            foreach (Hl7.Fhir.Model.DomainResource resource in resources)
+            {
+                foreach (DataType partialDateChild in resource.Children.Where(child => child.GetType().IsSubclassOf(typeof(DataType))))
+                {
+                    List<Extension> partialDateExtensions = partialDateChild.Extension.Where(ext => ext.Url.Equals(ExtensionURL.PartialDate)).ToList();
+                    foreach (Extension partialDateExtension in partialDateExtensions)
+                    {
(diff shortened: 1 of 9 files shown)


VRDR/DeathRecord_util.cs
```diff
@@ -1075,6 +1075,56 @@ public static void updateGivenHumanName(string[] value, List<HumanName> names)
...
+        /// <summary>Helper method to validate that all PartialDate and PartialDateTime exensions are valid and have the valid required sub-extensions.</summary>
+        /// <param name="bundle">The bundle in which to validate the PartialDate/Time extensions.</param>
+        private static void ValidatePartialDates(Bundle bundle)
+        {
+            System.Text.StringBuilder errors = new System.Text.StringBuilder();
+            List<Resource> resources = bundle.Entry.Select(entry => entry.Resource).ToList();
+
+            foreach (Hl7.Fhir.Model.Resource resource in resources)
+            {
+                foreach (DataType child in resource.Children.Where(child => child.GetType().IsSubclassOf(typeof(DataType))))
```
RobertScalfani, line 1088: I don't love the use of `.IsSubclassOf` here. All it needs is to be filtered and cast to the highest-level type that supports Extensions, is there a better way to do this that the FHIR library supports?

whichwit:
Please kindly update PR title to follow [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format so mainline updates can be automatically captured in the change log. The PR description can optionally be included in the commit message but not required.

johnarleyburns:
Currently I'm fixing some partial date and partial datetime and partial time issues in testing with Utah, once these are resolved I will treat this PR as a priority for inclusion in the mainline.

VRDR/DeathRecord_util.cs
johnarleyburns: Not sure that there's a better way to do this.

johnarleyburns (approved):

PR #494, opened 2023-07-18
fix: handling of missing decedents last name LNAME

consultantusa:


johnarleyburns (changes requested):
Please change Assert.Equals(null to Assert.Null and add documentation to README about LNAME handling

saulakravitz (commented):
I suggest adding a failing round trip (FHIR->IJE->FHIR) test where FamilyName = "UNKNOWN".
When this is round tripped through IJE, the resulting FamilyName will be null.
This is expected, but I think documenting this issue with a test will make it easier to remember this quirk.

consultantusa:
> Looking good, please add one more roundtrip unit test, this one comes from Saul's request:
> 
> ```
>     record.FamilyName = "UNKNOWN";
>     ije = new IJEMortality(record, false);
>     DeathRecord record1 = ije.ToDeathRecord();
>     Assert.Null(record1.FamilyName); // UNKNOWN is special value, so it becomes a null when round tripped
> ```

Implemented, but extra fix was needed to accommodate the special case.

johnarleyburns (approved):

PR #549, opened 2023-12-21
fix: add System.Net.Http namespace reference

whichwit:
Need to add specific namespace reference to `System.Net.Http`. Without it test suite fails.


wswarr (approved):
Looks good

PR #607, opened 2024-08-19
feat: add message header validation

smacadam:
Move message header validation from the API to the library


whichwit:
- Add testing
- Use standalone exception class
- Fix check on death year
@smacadam

whichwit (approved):

PR #616, opened 2024-10-24
fix: check for well formed source before parsing

whichwit:
- [x] Check for JSON source
- [x] Check for XML source


smacadam:
Perfect, this should also be updated in the monorepo in the VitalRecord.CommonMessage. I can make the changes in that project to match what you have here.

smacadam (approved):

smacadam (approved):

PR #635, opened 2025-04-09
feat: reimplement STU2-STU3 conversion from vital-records-dotnet in vrdr-dotnet

saulakravitz:


whichwit:
Some initial thoughts:
- Users of STU2, with it being the earlier version, is unlikely to convert to STU3 in an efficient or meaningful way. The value of the tool may be best amplified in the context of STU3 implementations i.e. the monorepo.
- The method of the conversion relies a lot on text matching which will require frequent maintenance as the standard develops. Have we considered a round-trip approach using shared/common profile attributes?

saulakravitz:
- The utility of the conversion tool is during the period of time when both standards are in use.
- Once established as the standard, the need for this conversion would disappear, as would the need for the vrdr .NET library.
- There are no plans for changes to the VRDR STU3 IG.

PR #638, opened 2026-04-06
fix: replace status flag update based on the event uri and the destin…

Veroniquecdc:
generate replace status flag based on the event uri and the destinations endpoints.


sathupgit (commented):
Looks good.

PR #640, opened 2026-07-30
feat: format zip code and add a replacestatusflag class

Veroniquecdc:
Format the zip code to remove the hyphen
Add a new class ReplaceStatusFlag to use the messageheader information to compute the replace Status Flag.


sathupgit (approved):
Looks good

